PR #19, opened 2026-05-13
Remove presence on pagehide via fetch keepalive

### LINKED ISSUE #15: Stale presence when user leaves a site
I found that if someone is in the admin and then navigates to another site, then their presence continues to be incorrectly reported on the page they left. I think there needs to be a `pagehide` event which fires off a [beacon](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon) to register that their presence is no longer active on a given screen. (NB: A user may have the same page open in multiple tabs.)

Similarly, there should be a `pageshow` event which re-establishes their presence on the page. Note that this fires when a page is first loaded  or restored from bfcache, so it would address the issue in https://github.com/WordPress/presence-api/issues/14 by not waiting for heartbeat to tick.

josephfusco:
Closes #15.

Removes presence entries on `pagehide` instead of waiting for the TTL.

## Changes

- **pagehide DELETE** — `fetch keepalive` removes each `(room, client_id)` the page occupies. URL built with `URL` + `URLSearchParams`.
- **Server-side write** — `wp_set_presence()` runs during page render so the new page is established before the response leaves.
- **Heartbeat triggers** — `DOMContentLoaded`, `pageshow.persisted`, and `visibilitychange === 'visible'` each call `wp.heartbeat.connectNow()`.
- **Hidden-tab guard** — one early-return in the consolidated `heartbeat-send` handler suppresses both `presence-ping` and `presence-editor-ping`.
- **Frontend tracking** expanded from `is_singular()` posts to any URL via `wp_get_document_title()`.
- **Cleanup** — editor ping handler consolidated into the main one; stored `post_type` dropped (derivable from `post_id`); inline JS modernized (`const`/`let`, optional chaining, optional catch, indented heredoc); config emitted with `JSON_HEX_TAG | JSON_UNESCAPED_SLASHES`.

## Limitations

- Client IDs are per-user, not per-tab. Closing one of multiple same-room tabs briefly removes the shared row; the next heartbeat restores it.
- `wp-refresh-post-lock` writes its own row independently; backgrounded post-edit tabs continue to refresh it. Tracked in #21.

## Testing

With `wp-env`:

- Close a tab → user disappears from Who's Online within seconds, not at TTL.
- Navigate between admin pages → no flicker for other watchers.
- Visit `/`, an archive, a search, and a 404 → each row carries a non-empty `title`.
- Hide a tab → presence stops refreshing. Show it → resumes within milliseconds.

### DIFF
--- a/includes/heartbeat.php
+++ b/includes/heartbeat.php
@@ -34,16 +34,31 @@ function wp_presence_enqueue_heartbeat_ping() {
 		),
 	);
 
-	// On frontend singular views, pass the current post context to the heartbeat ping.
+	// Carry a title for any frontend URL so non-singular views (archives, search,
+	// the front page, taxonomies, 404s) are labeled too. is_singular() pages also
+	// include the post id/type so the Active Posts widget can group by post.
 	$front_context = null;
-	if ( ! is_admin() && is_singular() ) {
-		$queried = get_queried_object();
-		if ( $queried instanceof WP_Post ) {
-			$front_context = array(
-				'postId'   => $queried->ID,
-				'postType' => $queried->post_type,
-				'title'    => get_the_title( $queried ),
-			);
+	if ( ! is_admin() ) {
+		if ( is_front_page() ) {
+			$title = __( 'Home', 'presence-api' );
+		} else {
+			$strip_branding = static function ( $parts ) {
+				unset( $parts['tagline'], $parts['site'] );
+				return $parts;
+			};
+			add_filter( 'document_title_parts', $strip_branding );
+			$title = wp_get_document_title();
+			remove_filter( 'document_title_parts', $strip_branding );
+		}
+
+		$front_context = array( 'title' => $title );
+
+		if ( is_singular() ) {
+			$queried = get_queried_object();
+			if ( $queried instanceof WP_Post ) {
+				$front_context['postId']   = $queried->ID;
+				$front_context['postType'] = $queried->post_type;
+			}
 		}
 	}
 
@@ -79,9 +94,13 @@ function wp_presence_enqueue_heartbeat_ping() {
 
 	$admin_state = array( 'screen' => $screen_id );
 	if ( $front_context ) {
-		$admin_state['post_id']   = $front_context['postId'];
-		$admin_state['post_type'] = $front_context['postType'];
-		$admin_state['title']     = $front_context['title'];
+		if ( ! empty( $front_context['title'] ) ) {
+			$admin_state['title'] = $front_context['title'];
+		}
+		if ( ! empty( $front_context['postId'] ) ) {
+			$admin_state['post_id']   = $front_context['postId'];
+			$admin_state['post_type'] = $front_context['postType'];
+		}
 	}
 	wp_set_presence( 'admin/online', 'user-' . $user_id, $admin_state, $user_id );
 
@@ -134,10 +153,14 @@ function wp_presence_enqueue_heartbeat_ping() {
 
 			$(document).on('heartbeat-send', function (event, data) {
 				const ping = { screen: window.pagenow || 'front' };
-				if (frontContext && frontContext.postId) {
-					ping.post_id = frontContext.postId;
-					ping.post_type = frontContext.postType;
-					ping.title = frontContext.title;
+				if (frontContext) {
+					if (frontContext.title) {
+						ping.title = frontContext.title;
+					}
+					if (frontContext.postId) {
+						ping.post_id = frontContext.postId;
+						ping.post_type = frontContext.postType;
+					}
 				}
 				data['presence-ping'] = ping;
 
--- a/includes/heartbeat.php
+++ b/includes/heartbeat.php
@@ -34,9 +34,9 @@ function wp_presence_enqueue_heartbeat_ping() {
 		),
 	);
 
-	// Carry a title for any frontend URL so non-singular views (archives, search,
-	// the front page, taxonomies, 404s) are labeled too. is_singular() pages also
-	// include the post id/type so the Active Posts widget can group by post.
+	// Carry a title for any frontend URL so it shows up in the Who's Online
+	// widget (non-singular views — archives, search, the front page, taxonomies,
+	// 404s — are labeled too). is_singular() pages also carry the post id.
 	$front_context = null;
 	if ( ! is_admin() ) {
 		if ( is_front_page() ) {
@@ -56,8 +56,7 @@ function wp_presence_enqueue_heartbeat_ping() {
 		if ( is_singular() ) {
 			$queried = get_queried_object();
 			if ( $queried instanceof WP_Post ) {
-				$front_context['postId']   = $queried->ID;
-				$front_context['postType'] = $queried->post_type;
+				$front_context['postId'] = $queried->ID;
 			}
 		}
 	}
@@ -98,8 +97,7 @@ function wp_presence_enqueue_heartbeat_ping() {
 			$admin_state['title'] = $front_context['title'];
 		}
 		if ( ! empty( $front_context['postId'] ) ) {
-			$admin_state['post_id']   = $front_context['postId'];
-			$admin_state['post_type'] = $front_context['postType'];
+			$admin_state['post_id'] = $front_context['postId'];
 		}
 	}
 	wp_set_presence( 'admin/online', 'user-' . $user_id, $admin_state, $user_id );
@@ -159,7 +157,6 @@ function wp_presence_enqueue_heartbeat_ping() {
 					}
 					if (frontContext.postId) {
 						ping.post_id = frontContext.postId;
-						ping.post_type = frontContext.postType;
 					}
 				}
 				data['presence-ping'] = ping;
--- a/includes/heartbeat.php
+++ b/includes/heartbeat.php
@@ -130,12 +130,15 @@ function wp_presence_enqueue_heartbeat_ping() {
 			return;
 		}
 
+		// rest_url() already contains ?rest_route= on plain-permalink sites.
+		var separator = restUrl.indexOf('?') === -1 ? '?' : '&';
+
 		entries.forEach(function (entry) {
 			if (!entry || !entry.room || !entry.client_id) {
 				return;
 			}
 			var url = restUrl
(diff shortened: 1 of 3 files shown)


includes/heartbeat.php
```diff
@@ -24,69 +29,141 @@ function wp_presence_enqueue_heartbeat_ping() {
...
+		if (typeof window.fetch !== 'function') {
+			return;
+		}
+
+		entries.forEach(function (entry) {
+			if (!entry || !entry.room || !entry.client_id) {
+				return;
+			}
+			var url = restUrl
+				+ '?room=' + encodeURIComponent(entry.room)
```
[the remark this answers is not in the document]
josephfusco: Good catch — fixed in c879fe8. The DELETE URL now picks `?` or `&` depending on whether `restUrl` already contains a query string (which it does on plain-permalink sites).

includes/heartbeat.php
```diff
@@ -24,69 +29,141 @@ function wp_presence_enqueue_heartbeat_ping() {
...
-			if (typeof wp === "undefined" || typeof wp.heartbeat === "undefined") { return; }
-			$(document).on("heartbeat-send", function(event, data) {
-				data["presence-editor-ping"] = { post_id: %d };
-			});
-		})(jQuery);',
-			$post->ID
-		)
+	window.addEventListener('pagehide', function () {
+		leave();
+	});
```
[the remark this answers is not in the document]
josephfusco: Addressed in 741041d — went with the server-side-write option you suggested. wp_presence_enqueue_heartbeat_ping() now writes the admin/online entry (and the editor entry on post-edit screens) synchronously during the new page's render, so the gap between the old page's DELETE and the new page being re-established shrinks from "DELETE + heartbeat round-trip" to just server render time. The immediate connectNow() (23577e8) and subsequent heartbeats keep the entry refreshed after that.

includes/heartbeat.php
```diff
@@ -24,69 +24,182 @@ function wp_presence_enqueue_heartbeat_ping() {
...
 		$queried = get_queried_object();
 		if ( $queried instanceof WP_Post ) {
-			$front_context = wp_json_encode(
+			$front_context = array(
+				'postId'   => $queried->ID,
+				'postType' => $queried->post_type,
+				'title'    => get_the_title( $queried ),
+			);
+		}
+	}
```
westonruter: What about for non-singular queries? It seems like you could support this for any arbitrary URL via `wp_get_document_title()`.

```php
if ( is_front_page() ) {
    $title = __( 'Home' );
} else {
    $unset_fields = function ( $parts ) {
        unset( $parts['tagline'], $parts['site'] );
        return $parts;
    }
    add_filter( 'document_title_parts', $unset_fields );
    $title = wp_get_document_title();
    remove_filter( 'document_title_parts', $unset_fields );
}
```

includes/heartbeat.php
```diff
@@ -24,69 +24,182 @@ function wp_presence_enqueue_heartbeat_ping() {
...
-					ping.post_id = frontContext.postId;
-					ping.post_type = frontContext.postType;
-					ping.title = frontContext.title;
-				}
-				data["presence-ping"] = ping;
-			});
-		})(jQuery);',
-			$front_context
-		)
+		'window.wpPresenceConfig = ' . wp_json_encode( $config ) . ';',
```
westonruter: For safety:

```suggestion
		'window.wpPresenceConfig = ' . wp_json_encode( $config, JSON_HEX_TAG | JSON_UNESCAPED_SLASHES ) . ';',
```

westonruter: Alternatively:


```suggestion
		sprintf( 'window.wpPresenceConfig = %s;', wp_json_encode( $config, JSON_HEX_TAG | JSON_UNESCAPED_SLASHES ) ),
```

includes/heartbeat.php
```diff
@@ -24,69 +24,182 @@ function wp_presence_enqueue_heartbeat_ping() {
...
+	// bfcache restore: DOMContentLoaded won't fire.
+	window.addEventListener('pageshow', function (event) {
+		if (event.persisted) { tickNow(); }
+	});
+
+	window.addEventListener('pagehide', function () {
+		leave();
+	});
+})(jQuery);
+JS
```
westonruter: Let's indent all lines of this JS. With PHP 7.4 being the minimum version, we can use flexible heredoc/nowdoc syntax which allows indentation (and strips it from the stored string).

includes/heartbeat.php
```diff
@@ -24,69 +24,182 @@ function wp_presence_enqueue_heartbeat_ping() {
...
 		return;
 	}
 
-	$post = get_post();
+	var config = window.wpPresenceConfig || {};
+	var entries = Array.isArray(config.entries) ? config.entries : [];
+	var frontContext = config.frontContext || null;
+	var editorPostId = parseInt(config.editorPostId, 10) || 0;
+	var restUrl = config.restUrl || '';
+	var nonce = config.nonce || '';
```
westonruter: Let's use modern JS since we now can, so `const` instead of `var`.

includes/heartbeat.php
```diff
@@ -24,69 +24,182 @@ function wp_presence_enqueue_heartbeat_ping() {
...
+		// rest_url() already contains ?rest_route= on plain-permalink sites.
+		var separator = restUrl.indexOf('?') === -1 ? '?' : '&';
+
+		entries.forEach(function (entry) {
+			if (!entry || !entry.room || !entry.client_id) {
+				return;
+			}
+			var url = restUrl
+				+ separator + 'room=' + encodeURIComponent(entry.room)
+				+ '&client_id=' + encodeURIComponent(entry.client_id);
```
westonruter: This should use `URL` and the params added via `URLSearchParams`.

includes/heartbeat.php
```diff
@@ -24,69 +24,182 @@ function wp_presence_enqueue_heartbeat_ping() {
...
+			try {
+				window.fetch(url, {
+					method: 'DELETE',
+					credentials: 'same-origin',
+					keepalive: true,
+					headers: { 'X-WP-Nonce': nonce }
+				});
+			} catch (e) {
```
westonruter:
```suggestion
			} catch {
```

includes/heartbeat.php
josephfusco: Adopted in f50d14d - `wp_get_document_title()` with your document_title_parts filter for non-singular URLs, hardcoded Home for `is_front_page()`.  Singular views still also carry post_id/post_type.  Broadened the Who's Online widget handler to record title-only payloads.

josephfusco:
@westonruter Thanks for the initial review! Suggestions have been applied.

includes/widgets/class-wp-presence-widget-whos-online.php
```diff
@@ -724,13 +724,20 @@ public static function heartbeat_received( $response, $data, $screen_id ) { // p
...
+		// can group by post.
+		if ( 'front' === $screen ) {
+			if ( ! empty( $data['presence-ping']['title'] ) ) {
+				$state['title'] = sanitize_text_field( $data['presence-ping']['title'] );
+			}
+			if ( ! empty( $data['presence-ping']['post_id'] ) ) {
+				$front_post_id = absint( $data['presence-ping']['post_id'] );
+				if ( $front_post_id ) {
+					$state['post_id']   = $front_post_id;
+					$state['post_type'] = isset( $data['presence-ping']['post_type'] ) ? sanitize_key( $data['presence-ping']['post_type'] ) : 'post';
```
westonruter: To ensure `post_id` isn't negative and is for a real post:

```suggestion
			$post_id = (int) ( $data['presence-ping']['post_id'] ?? 0 );
			if ( $post_id > 0 ) {
				$front_post = get_post( $post );
				if ( $front_post ) {
					$state['post_id']   = $front_post->ID;
					$state['post_type'] = $front_post->post_type;
```

Aside: why store the `post_type`? Isn't the ID sufficient?

includes/heartbeat.php
```diff
@@ -24,69 +24,204 @@ function wp_presence_enqueue_heartbeat_ping() {
...
+			remove_filter( 'document_title_parts', $strip_branding );
+		}
+
+		$front_context = array( 'title' => $title );
+
+		if ( is_singular() ) {
+			$queried = get_queried_object();
+			if ( $queried instanceof WP_Post ) {
+				$front_context['postId']   = $queried->ID;
+				$front_context['postType'] = $queried->post_type;
```
westonruter: Why are the keys camelCase here but snake_case above?

includes/heartbeat.php
```diff
@@ -24,69 +24,204 @@ function wp_presence_enqueue_heartbeat_ping() {
...
-			'(function($) {
-			if (typeof wp === "undefined" || typeof wp.heartbeat === "undefined") { return; }
-			var frontContext = %s;
-			$(document).on("heartbeat-send", function(event, data) {
-				var ping = { screen: window.pagenow || "front" };
-				if (frontContext.postId) {
-					ping.post_id = frontContext.postId;
-					ping.post_type = frontContext.postType;
-					ping.title = frontContext.title;
+		<<<'JS'
```
westonruter, line 136: Aside: It would be better of this JS could be located in an external JS file so it can get ESLint, TypeScript, and other static analysis. It can then get pulled in here via `file_get_contents()` or else enqueued as a regular script.

includes/heartbeat.php
```diff
@@ -24,69 +24,204 @@ function wp_presence_enqueue_heartbeat_ping() {
...
+		: 'front';
+
+	$admin_state = array( 'screen' => $screen_id );
+	if ( $front_context ) {
+		if ( ! empty( $front_context['title'] ) ) {
+			$admin_state['title'] = $front_context['title'];
+		}
+		if ( ! empty( $front_context['postId'] ) ) {
+			$admin_state['post_id']   = $front_context['postId'];
+			$admin_state['post_type'] = $front_context['postType'];
```
westonruter: Per above, is the `post_type` really needing to be saved? The ID should allow that to be easily obtained.

includes/heartbeat.php
josephfusco: @westonruter Good call, will break that out into a separate PR. Thanks!

PR #20, opened 2026-05-13
Skip presence ping for hidden tabs and refresh on visibility

### LINKED ISSUE #16: Page visibility should be factored into presence
It also may make sense to take some action in response to `visibilitychange` events. A user who has a given page open in a background tab should not be shown as present as someone who is actively working on a tab. A beacon should probably fire whenever a `visibilitychange` event occurs, similar to the `pageshow` and `pagehide` events in #15 (cf. #14).

josephfusco:
Closes #16.

A user with a background tab on a presence-enabled page was still reported as actively present.

## Change

Two small additions to the inline heartbeat script:

1. In the `heartbeat-send` handler, return early when `document.visibilityState === 'hidden'`. The presence-ping is omitted, and the existing 60 s TTL plus the cron sweep clean up the stale entry naturally.
2. On `visibilitychange` to `'visible'`, call `wp.heartbeat.connectNow()` so the user is re-established in the room within milliseconds rather than waiting for the next heartbeat interval.

This is intentionally a soft fix (no explicit `DELETE`) so it can land independently of any pagehide-based cleanup. If we later add pagehide DELETEs (see #15 / #19), a small follow-up could fire the same `leave()` on hidden as well; out of scope here to keep this diff minimal.

## Files

- `includes/heartbeat.php`

## Verification

With wp-env running, log in as editor in window A and as admin in window B. From the admin window, watch the Who's Online widget. In window A, switch to a different tab — within roughly a heartbeat interval the editor's screen should stop updating. Switch back to the admin tab and the editor's presence should refresh on the next tick.

### DIFF
--- a/includes/heartbeat.php
+++ b/includes/heartbeat.php
@@ -47,7 +47,7 @@ function wp_presence_enqueue_heartbeat_ping() {
 			var frontContext = %s;
 			$(document).on("heartbeat-send", function(event, data) {
 				// A user with a hidden tab should not be reported as actively present.
-				// Skip the ping so the existing entry expires via the 60s TTL.
+				// Skip the ping so the existing entry expires via the default TTL.
 				if (document.visibilityState === "hidden") {
 					return;
 				}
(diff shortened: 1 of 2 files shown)


includes/heartbeat.php
```diff
@@ -46,6 +46,11 @@ function wp_presence_enqueue_heartbeat_ping() {
 			if (typeof wp === "undefined" || typeof wp.heartbeat === "undefined") { return; }
 			var frontContext = %s;
 			$(document).on("heartbeat-send", function(event, data) {
+				// A user with a hidden tab should not be reported as actively present.
+				// Skip the ping so the existing entry expires via the 60s TTL.
+				if (typeof document.visibilityState === "string" && document.visibilityState === "hidden") {
```
westonruter: I don't think the `typeof` is needed. It's supported by all browsers WP supports: https://caniuse.com/pagevisibility


```suggestion
				if (document.visibilityState === "hidden") {
```

westonruter (approved):

PR #30, opened 2026-06-10
Add a composer audit CI workflow

josephfusco:
Fails CI when any direct or transitive dependency in `composer.lock` has a known security advisory. Runs on every PR that touches the lockfile, plus weekly so newly-disclosed advisories against unchanged dependencies still trip CI.


josephfusco:
Closing — redundant with Dependabot, which re-detects the same vulns and adds the auto-fix PR. Replacing with a `.github/dependabot.yml` in a follow-up.

Issue #21, opened 2026-05-14
Hidden post-edit tabs still refresh the post-lock presence entry


josephfusco:
Found while reviewing #20 (Page Visibility integration).

The visibility guard introduced in #20 suppresses our own `presence-ping` and `presence-editor-ping` while a tab is hidden, but the **post-lock** presence entry (`lock-{user_id}` in the per-post room) is written by `wp_presence_bridge_post_lock()` off WP core's `wp-refresh-post-lock` heartbeat. That ping is set by core's edit-lock JS, not by our inline script, so we can't suppress it from the same place.

Result: a backgrounded post-edit tab still keeps the `lock-{id}` row alive in `postType/{type}:{id}` even after `editor-{id}` has aged out.

Whether this matters is a product call:
- If post locks should outlive a brief glance-away → leave as-is.
- If we want strict "hidden = inactive" → bridge needs its own visibility check, or core's edit-lock JS needs upstream cooperation.

Likely fix: in `includes/post-lock-bridge.php::wp_presence_bridge_post_lock()`, accept (or expose) a visibility signal — easiest path is to have the inline script null out `data["wp-refresh-post-lock"]` when hidden, so core's handler runs to completion without us writing the bridge entry.

cc / follow-up to #20.

josephfusco:
Fixed in #78, merged 2026-07-27.

The fix took the approach suggested here: the ping script drops core's post-lock payload while the document is hidden, so `wp_presence_bridge_post_lock()` never writes the `lock-{user_id}` entry and it ages out on the default TTL. `assets/js/presence-ping.js`:

```js
if (document.visibilityState === 'hidden') {
	delete data['wp-refresh-post-lock'];
	return;
}
```

This did not close automatically because the PR body wrote `Resolves Issue: #21` — GitHub only parses a closing keyword when the reference follows it directly, so the interposed word breaks the link.

Note that #134 lists this issue as something it would resolve. That is no longer a dependency: #21 has its own fix, and #134 remains open on its own merits.

PR #78, opened 2026-07-17
Fix: Hidden post-edit tabs still refresh the post-lock presence entry

Intenzi:
Resolves Issue: #21 
Related PRs: #19 , #20 

## How?
- Deletes the `wp-refresh-post-lock` key from the shared heartbeat `data` object when the tab/document is backgrounded (`document.visibilityState === 'hidden'`).
  - Because this key is deleted, WordPress Core's own heartbeat ping doesn't get sent and is thus no longer causing the intercept to update the presence database.
- Wrapped the heartbeat listener in a document ready hook (`$(function() { ... })`). Because our script runs earlier/faster than WordPress Core, registering our listener resulted in an empty data object when attempting to delete the key. Delaying to document ready ensures that the `data['wp-refresh-post-lock']` key exists by the time our handler runs, allowing us to successfully delete it.
- Updated the visibility E2E test to assert that `wp-refresh-post-lock` is omitted from the heartbeat payload when the editor tab is hidden.

> [!NOTE]
> Since we defer the registration specifically to ensure we run after WordPress Core's handler. Keeping all of our code inside this single deferred handler is the simplest approach for now. If we ever need to run other tasks at the very beginning of the heartbeat tick in the future, we can easily split this into two separate listeners (one immediate, one deferred) without any breaking changes.

## Use of AI Tools

AI assistance: Yes
Tool(s): Antigravity IDE
Model(s): Gemini 3.5 Flash (Low)
Used for: Identifying relevant code file, going through related PRs, it suggested utilising the deferred execution later when debugging.

### DIFF
--- a/tests/e2e/presence-visibility.test.js
+++ b/tests/e2e/presence-visibility.test.js
@@ -89,7 +89,7 @@ test.describe( 'Presence Visibility', () => {
 		expect( visible[ 'presence-ping' ].screen ).toBeTruthy();
 	} );
 
-	test( 'heartbeat-send omits presence-editor-ping while editor tab is hidden', async ( {
+	test( 'heartbeat-send omits presence-editor-ping and wp-refresh-post-lock while editor tab is hidden', async ( {
 		admin,
 		page,
 		requestUtils,
@@ -108,6 +108,7 @@ test.describe( 'Presence Visibility', () => {
 		await setVisibility( page, 'hidden' );
 		const hidden = await captureHeartbeatSend( page );
 		expect( hidden[ 'presence-editor-ping' ] ).toBeUndefined();
+		expect( hidden[ 'wp-refresh-post-lock' ] ).toBeUndefined();
 
 		await setVisibility( page, 'visible' );
 		const visible = await captureHeartbeatSend( page );
--- a/tests/e2e/presence-visibility.test.js
+++ b/tests/e2e/presence-visibility.test.js
@@ -16,7 +16,68 @@
  * @package WordPress
  * @since 7.1.0
  */
-import { test, expect } from '@wordpress/e2e-test-utils-playwright';
+import { test as base, expect } from '@wordpress/e2e-test-utils-playwright';
+import { chromium } from '@playwright/test';
+
+const BASE_URL = ( process.env.WP_BASE_URL || 'http://localhost:8888' ).replace( /\/$/, '' );
+
+const TEST_USERS = [
+	{
+		username: 'presence_test_b',
+		email: 'presence_test_b@example.com',
+		firstName: 'User',
+		lastName: 'B',
+		password: 'password',
+		roles: [ 'editor' ],
+	},
+];
+
+const test = base.extend( {
+	testUsers: [
+		async ( { requestUtils }, use ) => {
+			for ( const user of TEST_USERS ) {
+				await requestUtils.createUser( user ).catch( ( error ) => {
+					if ( error?.code !== 'existing_user_login' ) {
+						throw error;
+					}
+				} );
+			}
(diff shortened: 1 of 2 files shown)


tests/e2e/presence-visibility.test.js
```diff
@@ -108,6 +108,7 @@ test.describe( 'Presence Visibility', () => {
 		await setVisibility( page, 'hidden' );
 		const hidden = await captureHeartbeatSend( page );
 		expect( hidden[ 'presence-editor-ping' ] ).toBeUndefined();
+		expect( hidden[ 'wp-refresh-post-lock' ] ).toBeUndefined();
```
josephfusco, line 172: Might be worth adding a positive assertion on the visible side too, so a regression that always deletes the key doesn't slip through:

```suggestion
		expect( visible[ 'wp-refresh-post-lock' ] ).toBeDefined();

```

josephfusco (changes requested):
Good fix.  One suggestion inline.

tests/e2e/presence-visibility.test.js
Intenzi: Hi @josephfusco, I'd love your input on the implementation direction here.

I went through with implementing the positive assertion and it has lead to a blocker.
After applying the positive assertion, the test was consistently failing when I ran the e2e tests locally.

Upon digging deeper into how Core handles post locking, I realised that the current negative assertion implementation done by me is also a false positive. 

In Core's `post.js`, `wp-refresh-post-lock` is only attached if `#post-lock-dialog` exists in the DOM:

```javascript
// wp-admin/js/post.js
	/**
	 * Heartbeat locks.
	 *
	 * Used to lock editing of an object by only one user at a time.
	 *
	 * When the user does not send a heartbeat in a heartbeat-time
	 * the user is no longer editing and another user can start editing.
	 */
	$(document).on( 'heartbeat-send.refresh-lock', function( e, data ) {
		var lock = $('#active_post_lock').val(),
			post_id = $('#post_ID').val(),
			send = {};

		if ( ! post_id || ! $('#post-lock-dialog').length )
			return;

		send.post_id = post_id;

		if ( lock )
			send.lock = lock;

		data['wp-refresh-post-lock'] = send;

```

In standard Core PHP (`_admin_notice_post_locked()`), Core only outputs `#post-lock-dialog` into the HTML if there is an active lock conflict from _another user_. The another user check is performed inside `wp_check_post_lock` :-
```php
...
	if ( $time && $time > time() - $time_window && get_current_user_id() !== $user ) {
		return $user;
	}

	return false;
}
```

For a single user editing a post alone, Core never outputs that div, so `post.js` skips attaching `wp-refresh-post-lock` altogether.

Because of this, the test was passing the negative assertion simply because Core never attached the key in the first place, not because our visibility handler stripped it.

### Proposed Fix:
We can implement a separate test to use a genuine multi-user flow using Playwright's `browser.newContext()` to simulate two actual users (User A & User B):

User B opens the post first in Context 1 to acquire the lock.
User A opens the post second in Context 2, which triggers Core's multi-user lock dialog naturally.

Then we perform visibility, detached from visibility tests on it.

Alternatively, we could simulate the lock conflict state by updating `_edit_lock` meta prior to page load if we want a single-browser run.

Which approach would you prefer for the test suite?

josephfusco: Great catch, you're right that the negative assertion was always true. I'd maybe lean towards the `browser.newContext()` approach.  Have User B navigate to the post edit page first to naturally acquire the lock, then User A opens the same post and the lock dialog is present. Run the visibility assertions from there. That's the most authentic test of the behavior.

josephfusco (approved):

PR #86, opened 2026-07-22
ci: add dedicated assets-only deploy workflow for WordPress.org

josephfusco:
The main deploy workflow bundles assets with the trunk commit and aborts when the version is already published — making it impossible to update plugin directory assets independently. This adds a standalone workflow that talks directly to the SVN assets directory, bypassing the version check entirely.

Related: 

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): claude-sonnet-4-6
Used for: Drafting and implementation


josephfusco:
Superseded by #87.

PR #111, opened 2026-07-24
Use $user_id from wp_logout hook instead of get_current_user_id()

### LINKED ISSUE #110: wp_logout hook fails to clear presence data on logout
When a user logs out, their presence data is not removed from the database. They remain visible as online to other users until the default TTL expires naturally.

`wp_presence_on_logout()` in `includes/lifecycle.php` calls `get_current_user_id()` to determine whose presence to remove. However, by the time the `wp_logout` action fires in WordPress core, `wp_clear_auth_cookie()` has already run, clearing the auth cookie and causing it to return 0 for remainder of the request.

The capability guard then fails, the function exits early, and no presence entries are removed.

Expected Behavior - The logged-out user should be removed from all presence rooms immediately upon logout.

Proposed Fix - Since WordPress 5.5, the `wp_logout` action passes the logged-out user's ID as its first argument. `wp_presence_on_logout()` should accept and use this `$user_id` parameter instead of relying on `get_current_user_id()`.

AshishJii:
Fixes a bug where logging out fails to remove the user's presence data from the database.

Closes #110 

`get_current_user_id()` returns `0` by the time `wp_logout` fires because `wp_clear_auth_cookie()` has already run. The existing guard then short-circuits the function before `wp_remove_user_presence()` is ever called.

Since WordPress 5.5, `wp_logout` passes the logged-out user's ID as its first argument. This PR uses that argument directly. [reference](https://developer.wordpress.org/reference/hooks/wp_logout/)

### Changes

- **`includes/lifecycle.php`** — Updated `wp_presence_on_logout()` to accept `$user_id` as a parameter; removed the broken `get_current_user_id()` call.
- **`presence-api.php`** — Made the `add_action()` registration explicit with `10, 1` for readability.
- **`tests/test-lifecycle.php`** — Updated two existing logout tests to pass `$user_id` directly. Added a regression test that calls `wp_set_current_user( 0 )` before invoking the callback, explicitly simulating the real logout timing.

### Testing

```
OK (81 tests, 165 assertions)
```

### DIFF
--- a/tests/test-lifecycle.php
+++ b/tests/test-lifecycle.php
@@ -80,7 +80,7 @@ class WP_Test_Presence_Lifecycle extends WP_UnitTestCase {
 		wp_set_presence( 'admin/online', 'user-' . self::$editor_id, array(), self::$editor_id );
 		wp_set_presence( 'postType/post:1', 'lock-' . self::$editor_id, array(), self::$editor_id );
 
-		// Simulate real wp_logout timing
+		// Simulate real wp_logout timing.
 		// Auth cookie has been cleared, so get_current_user_id() would return 0.
 		wp_set_current_user( 0 );
 
--- a/tests/test-lifecycle.php
+++ b/tests/test-lifecycle.php
@@ -54,8 +54,7 @@ class WP_Test_Presence_Lifecycle extends WP_UnitTestCase {
 		wp_set_presence( 'admin/online', 'user-' . self::$editor_id, array(), self::$editor_id );
 		wp_set_presence( 'postType/post:1', 'lock-' . self::$editor_id, array(), self::$editor_id );
 
-		wp_set_current_user( self::$editor_id );
-		wp_presence_on_logout();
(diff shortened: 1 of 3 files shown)


tests/test-lifecycle.php
```diff
@@ -67,11 +66,26 @@ public function test_logout_skips_subscriber() {
...
 	}
+
+	/**
+	 * @covers ::wp_presence_on_logout
+	 */
+	public function test_logout_clears_presence_without_current_user() {
+		wp_set_presence( 'admin/online', 'user-' . self::$editor_id, array(), self::$editor_id );
+		wp_set_presence( 'postType/post:1', 'lock-' . self::$editor_id, array(), self::$editor_id );
+
+		// Simulate real wp_logout timing
```
josephfusco: Nit: missing period.

```suggestion
		// Simulate real wp_logout timing.
```

josephfusco (approved):
@AshishJii Thank you for these changes.  One small nit, otherwise this looks good to go!

Issue #129, opened 2026-07-26
Remove sanitize_text_field() from the REST data payload sanitizer


josephfusco:
`sanitize_data_param()` runs `sanitize_text_field()` on every string value in the `data` object before it is stored:

```php
} elseif ( is_string( $value ) ) {
    $sanitized[ $key ] = sanitize_text_field( $value );
}
```

`sanitize_text_field()` strips newlines and trims whitespace. It is designed for single-line HTML form inputs. Any caller storing a multiline string, a value with leading or trailing whitespace, or HTML entities that are semantically meaningful will have their data silently corrupted on write.

The XSS risk this appears to guard against is already handled correctly:

- `wp_json_encode()` escapes output at storage time
- The REST API encodes responses as JSON with appropriate headers
- `sanitize_text_field()` at write time does not add security for JSON-encoded data

**The fix:** remove the `sanitize_text_field()` call on string values. Replace it with an `is_string()` guard that passes the value through untouched. Integer, float, and boolean values are already passed through without transformation and that model is correct -- string values should follow the same pattern.

```php
} elseif ( is_string( $value ) ) {
    $sanitized[ $key ] = $value;
}
```

This is not "no sanitization" -- it is trusting output-time encoding, which is where the protection actually lives.

josephfusco:
This issue was identified as part of a peer review conducted by Claude Sonnet 4.6, facilitated by @josephfusco.

i-am-chitti:
@josephfusco I've raised a PR here - https://github.com/WordPress/presence-api/pull/145 for the fix.

josephfusco:
Fixed in #145, merged as 4f88543.

`sanitize_data_recursive()` now passes string values through untouched, exactly as proposed:

```php
} elseif ( is_string( $value ) ) {
	$sanitized[ $key ] = $value;
}
```

Keys are still run through `sanitize_text_field()`, which was out of scope here. Regression coverage is in `test_sanitize_data_does_not_strip_html_or_whitespace` (`tests/test-rest-controller.php`), asserting a multiline value survives the round trip intact.

PR #145, opened 2026-07-26
REST API: Remove value sanitization from arbitrary data parameter

i-am-chitti:
## Description

This PR fixes an issue where multiline strings, leading/trailing whitespace, and HTML tags are silently corrupted or stripped when saving the arbitrary `data` parameter in a presence payload.

Currently, `WP_REST_Presence_Controller::sanitize_data_recursive()` runs `sanitize_text_field()` on every string value in the `data` array. Since the REST API encodes responses as JSON, and all output contexts (like the dashboard widget) escape values on render, this write-time sanitization is redundant and destructive.

This change removes `sanitize_text_field()` on the values, passing strings through untouched (similar to integer, float, and boolean values), and updates documentation and unit tests.

Fixes 
- #129 

## Verification

* **PHPUnit tests**: Run `npx wp-env run tests-cli --env-cwd='wp-content/plugins/presence-api' -- ./vendor/bin/phpunit --testdox`

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Opus 4.8
Used for - investigating and identifying usage of the function, updating test cases

### DIFF
--- a/tests/test-rest-controller.php
+++ b/tests/test-rest-controller.php
@@ -227,6 +227,7 @@ class WP_Test_Presence_REST_Controller extends WP_UnitTestCase {
 
 	/**
 	 * @covers WP_REST_Presence_Controller::sanitize_data_param
+	 * @covers WP_REST_Presence_Controller::sanitize_data_recursive
 	 */
 	public function test_sanitize_data_does_not_strip_html_or_whitespace() {
 		$controller = new WP_REST_Presence_Controller();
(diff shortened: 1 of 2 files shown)


tests/test-rest-controller.php
```diff
@@ -228,14 +228,19 @@ public function test_get_rooms_returns_data() {
 	/**
 	 * @covers WP_REST_Presence_Controller::sanitize_data_param
```
josephfusco, line 229:
```suggestion
	 * @covers WP_REST_Presence_Controller::sanitize_data_param
	 * @covers WP_REST_Presence_Controller::sanitize_data_recursive
```

josephfusco (changes requested):
Hey @i-am-chitti, one small comment relating to coverage reporting.

josephfusco (approved):

josephfusco:
@i-am-chitti I went ahead and merged. Thank you for working on this!

PR #183, opened 2026-08-06
ci: bump codeql-action to v4.37.4 and group its Dependabot updates

josephfusco:
Dependabot split the v4.37.4 bump into #180 and #181, and neither passes on its own because CodeQL errors out when `init` and `analyze` are on different versions.

Supersedes #180 and #181.

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5
Used for: Assisting with diagnosis and implementation

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -24,3 +24,8 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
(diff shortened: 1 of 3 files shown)


josephfusco:
@copilot rebase

PR #213, opened 2026-08-08
fix: store presence data as longtext and compare schema version as an integer

### LINKED ISSUE #207: Schema version is a string compared with strict equality
`presence-api.php:49` defines the schema version as a string:

```php
define( 'WP_PRESENCE_DB_VERSION', '1' );
```

and `wp_maybe_create_presence_table()` gates on a strict comparison:

```php
if ( get_option( 'wp_presence_db_version' ) === WP_PRESENCE_DB_VERSION ) {
	return;
}
```

Core keeps `$wp_db_version` an integer and compares numerically. A strict comparison against an option is brittle because the value's type depends on where it came from: the database returns a string, but a persistent object cache returns whatever type was last written, and any future change to the constant's type breaks the check silently.

It fails in one of two directions, both quiet: the upgrade never runs, or `dbDelta()` runs on every admin request.

Fix is to make the constant an integer and cast on read, matching `wp_upgrade()`.

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

### LINKED ISSUE #208: Presence data column should be longtext, not text
The `data` column in the `wp_presence` schema (`includes/functions.php`, inside `wp_maybe_create_presence_table()`) is declared `text NOT NULL` and stores JSON.

Core uses `longtext` for columns holding arbitrary serialized payloads, `meta_value` being the obvious precedent. `text` caps at 64KB, and on a MySQL server not running in strict mode an oversized value is truncated silently, which for JSON means an unparseable payload rather than a visible error.

Presence state is small today, but the column is public API surface for anything storing state in a room, and the schema is far cheaper to widen now than after the core proposal.

Needs a `WP_PRESENCE_DB_VERSION` bump so `dbDelta()` picks up the column change.

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

aldorizona10-glitch:
Widen the `wp_presence.data` column from `text` to `longtext`, and make `WP_PRESENCE_DB_VERSION` an integer compared numerically.

### What & why
- **`data text` → `data longtext`** (#208): the column stores JSON. `text` caps at ~64KB and, on a non-strict MySQL server, truncates silently — turning an oversized value into unparseable JSON rather than a visible error. `longtext` matches core's precedent for arbitrary serialized payloads (`meta_value`).
- **Schema version bump `1` → `2`**: required so `dbDelta()` applies the column change on existing installs.
- **`WP_PRESENCE_DB_VERSION` `'1'` (string) → `2` (integer), cast on read** (#207): the old strict comparison `get_option( … ) === WP_PRESENCE_DB_VERSION` was brittle — the option's runtime type depends on its source (the DB returns a string; a persistent object cache returns whatever was last written), so the guard could silently skip the upgrade or run `dbDelta()` on every admin request. Casting on read matches `wp_upgrade()`.

Both issues touch the same constant and function, so they're fixed together to avoid a conflicting version bump.

Fixes #207
Fixes #208

### Testing
- The existing `tests/test-functions.php::test_schema_migration_on_version_bump` and `::test_schema_migration_skipped_when_current` remain valid: after migration the option is written as the integer constant and read back as the same int within the request, so the `assertSame` checks hold; the outdated-version path (`'0.0'`) now casts to `0 !== 2` and correctly triggers the upgrade.
- I did not have a local `wp-env` PHP/MySQL environment to run PHPUnit/PHPCS/PHPStan; CI (PHP 7.4 + 8.3, multisite, Playwright) is the authoritative check.

### Use of AI Tools
AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 4.8
Used for: Implementing the change, the commit message, and this description, following the fixes described by the issue authors in #207 and #208. I have reviewed the diff and take responsibility for it.


josephfusco:
@aldorizona10-glitch Thanks for the PR! Something wonky is going on with the props-bot, going to look at that now so we can get this merged in.

josephfusco:
@aldorizona10-glitch Two things needed to credit you properly:

1. **Your commit email isn't tied to your GitHub account**, so the props bot couldn't identify you and printed the raw address instead (sorry about that). Fix by adding and verifying it in [Settings → Emails](https://github.com/settings/emails), or by amending onto your `@users.noreply.github.com` address and force-pushing. Only the second option gets your address out of the public commit history.

2. **Link GitHub to WordPress.org**: [profiles.wordpress.org/me/profile/edit/](https://profiles.wordpress.org/me/profile/edit/) → Account & Security → "link your GitHub account". Create a WordPress.org account first if you don't have one.

Do 1 before 2, otherwise the bot still can't find you. Ping me when done and I'll re-run it.

Worth dropping the `Co-Authored-By: Claude` trailer while you're amending. You already disclosed the AI usage in the PR description, which is where we want it, so the trailer just duplicates it into commit metadata and into the props list. You can turn it off permanently with [Claude Code's attribution settings](https://code.claude.com/docs/en/settings#attribution-settings).

aldorizona10-glitch:
Thanks @josephfusco! Updated on my side:

- Switched my commit email to my GitHub `@users.noreply.github.com` address and force-pushed, so the raw address is out of the commit history.
- Dropped the `Co-Authored-By` trailer (AI use stays disclosed in the PR description).

Linking my WordPress.org account next and will confirm here so you can re-run props-bot. Thanks for catching the redaction bug (#223)!

josephfusco:
Synced your branch with main so it isn't behind 0.1.15. The only conflict was the two constants; I kept WP_PRESENCE_VERSION at main's 0.1.15 since release-please owns it, and your WP_PRESENCE_DB_VERSION as integer 2. Pull before you push again (if even needed, as this looks good to me and is ready to merge once a WordPress.org account is linked).

aldorizona10-glitch:
@josephfusco Both done now:

- Commit email is on my GitHub `@users.noreply.github.com` address (raw address removed from history).
- Linked my GitHub account (`aldorizona10-glitch`) to WordPress.org.

Ready for a props-bot re-run whenever you get a chance — thanks for shepherding this through and for the sync!

Issue #217, opened 2026-08-08
Narrow @covers annotations suppress real coverage; Active Posts widget reports 4.87% while being well tested


josephfusco:
`includes/widgets/class-wp-presence-widget-active-posts.php` reports 4.87% coverage, 117 lines missed. That reads like an untested file. It is not. `tests/widgets/test-widget-active-posts.php` has nine tests that exercise the widget's core logic in real depth, including capability filtering per post, idle versus active status, multi-user grouping, and non-post room exclusion.

The tests are not being credited, because every one of them carries a narrow `@covers`:

```php
/**
 * @covers WP_Presence_Widget_Active_Posts::heartbeat_received
 */
```

PHPUnit records coverage only for lines inside the annotated unit. It does not follow into callees. `heartbeat_received()` is a six-line pass-through that delegates everything to the private `build_active_posts_data()`:

https://github.com/WordPress/presence-api/blob/main/includes/widgets/class-wp-presence-widget-active-posts.php#L254-L266

The arithmetic confirms it exactly. That method's executable lines are 255, 256, 259, 260, 263, and 265, which is six lines. Codecov reports six covered lines out of 123. Every assertion those nine tests make about `build_active_posts_data()` scores zero.

**This is not limited to one file.** Nearly every test in the suite carries a narrow `@covers`:

| Test file | Tests | `@covers` |
|---|---|---|
| `tests/test-functions.php` | 32 | 33 |
| `tests/test-rest-controller.php` | 21 | 22 |
| `tests/test-screen-revisions.php` | 20 | 20 |
| `tests/widgets/test-widget-active-posts.php` | 9 | 9 |
| `tests/test-heartbeat.php` | 7 | 7 |

`includes/functions.php` is unaffected at 98.46%, because its tests target public functions that do their own work. The files that suffer are the ones that route a thin public entry point into private helpers, which is most of the admin and widget layer. Some part of the reported gap in `screen-revisions.php` (51.02%) and the REST controller (45.06%) is likely the same artefact rather than absent tests.

**Why this matters beyond the number**

A misleading coverage report is worse than a low one. It sends effort at files that are already tested and hides files that are not. Any plan to raise coverage has to correct this first, or it will be planned against fiction.

**Suggested direction**

Extend the annotations to name the helpers the test actually drives:

```php
/**
 * @covers WP_Presence_Widget_Active_Posts::heartbeat_received
 * @covers WP_Presence_Widget_Active_Posts::build_active_posts_data
 */
```

[...]

PR #234, opened 2026-08-08
test: cover empty-result and deleted-user branches in functions.php

### LINKED ISSUE #224: Two untested branches remain in functions.php after the @covers widening
Widening the `@covers` annotations (#217) took `includes/functions.php` to 212/214 statements. The two remaining lines are genuine test gaps, not annotation artifacts.

**`includes/functions.php:258`** in `wp_get_user_presence()`:

```php
if ( ! $results ) {
    return array();
}
```

No test calls `wp_get_user_presence()` for a user with no live presence rows, so the empty-result early return never executes.

**`includes/functions.php:502`** in `wp_get_active_rooms()`:

```php
$user = get_userdata( $uid );

if ( ! $user ) {
    continue;
}
```

No test drives `wp_get_active_rooms()` with a presence row whose `user_id` points at a deleted user, so the skip branch never executes. This is the same deleted-user condition the REST controller already has a test for, just not at the function level.

**Suggested fix**

Two additions to `tests/test-functions.php`:

1. Assert `wp_get_user_presence()` returns an empty array for a user with no rows.
2. Insert a presence row, `wp_delete_user()` the author, then assert `wp_get_active_rooms()` omits them from the user list rather than emitting an entry with empty data.

Both are self-contained and need no new fixtures beyond the existing factory users.

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

i-am-chitti:
### Summary

Two branches in `includes/functions.php` were never exercised by the suite after the `@covers` widening in #217:

Adds one test for each, following the existing patterns in `tests/test-functions.php` (including the multisite-aware `wp_delete_user()`/`wpmu_delete_user()` branch already used in `tests/test-rest-controller.php`).

Fixes #224

### Testing
- `phpcs`
- `npm run test`

### DIFF
--- a/tests/test-functions.php
+++ b/tests/test-functions.php
@@ -144,6 +144,13 @@ class WP_Test_Presence_Functions extends WP_UnitTestCase {
 	}
 
 	/**
+	 * @covers ::wp_get_user_presence
+	 */
+	public function test_get_user_presence_returns_empty_array_when_none_present() {
+		$this->assertSame( array(), wp_get_user_presence( self::$subscriber_id ) );
+	}
+
+	/**
 	 * @covers ::wp_can_access_presence_room
 	 */
 	public function test_editor_can_access_room() {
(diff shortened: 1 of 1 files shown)


tests/test-functions.php
```diff
@@ -143,6 +143,13 @@ public function test_get_user_presence_across_rooms() {
 		$this->assertCount( 2, $entries );
 	}
 
+	/**
+	 * @covers ::wp_get_user_presence
+	 */
+	public function test_get_user_presence_returns_empty_array_when_none_present() {
+		$this->assertSame( array(), wp_get_user_presence( self::$subscriber_id ) );
+	}
```
josephfusco, line 154: Nit: tear_down() empties the table, so this passes because there are no rows at all. Adding a row for a different user first makes it also check that the user filter works:

```php
wp_set_presence( 'test/room', 'client-1', array(), self::$editor_id );
```

tests/test-functions.php
```diff
@@ -369,6 +376,27 @@ public function test_get_active_rooms_empty() {
...
+		} else {
+			wp_delete_user( $temp_user_id );
+		}
+
+		$rooms = wp_get_active_rooms();
+
+		$this->assertCount( 1, $rooms );
+		$this->assertSame( 0, $rooms[0]['user_count'] );
+		$this->assertSame( array(), $rooms[0]['users'] );
+	}
```
josephfusco, line 402: Nit: Same idea here. Adding a second user who isn't deleted would show that the deleted one gets skipped while the other still shows up:

```php
wp_set_presence( 'admin/online', 'client-live', array(), self::$editor_id );
```

Then user_count would be 1 and users would hold the editor, instead of 0 and empty.

josephfusco (approved):
Confirmed this closes #224. Thanks Deepak!

i-am-chitti:
Ah, good catch - both fixed. Added a live counterpart in each test so they actually prove the filtering/skip logic instead of just working against an empty table

PR #240, opened 2026-08-08
test: add coverage for cron scheduling

### LINKED ISSUE #220: Cron scheduling has no tests (0% coverage)
`includes/cron.php` reports 0% coverage, 6 lines missed. Small, but it is the scheduling half of the cleanup mechanism and nothing asserts it.

Part of the zero is the `@covers` artefact from #217. `wp_presence_schedule_cleanup()` does run during the suite, called from `wp_presence_activate()` in `tests/test-table-creation.php::test_activation_provisions_the_current_site`, but that test is annotated `@covers ::wp_presence_activate`, so the cron function gets no credit. It is also called from `tests/bootstrap.php`, which PHPUnit does not count because coverage collection starts after bootstrap.

`wp_presence_cron_schedules()` is genuinely untested.

**Suggested direction**

A small `tests/test-cron.php`:

- `wp_presence_cron_schedules()` adds `wp_presence_every_minute` with a 60 second interval and a translated display string, and preserves the schedules already in the array
- The filter is actually registered, so `wp_get_schedules()` includes the interval
- `wp_presence_schedule_cleanup()` schedules `wp_delete_expired_presence_data` when nothing is scheduled
- Calling it twice does not double schedule, which is the guard on line 16

If a future DRY pass adds `wp_presence_unschedule_cleanup()`, cover that here too.

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

aldorizona10-glitch:
Adds `tests/test-cron.php` covering `includes/cron.php`, following the suggested direction in #220.

### Covers
- `wp_presence_cron_schedules()` registers `wp_presence_every_minute` (60s interval + non-empty display) and preserves the schedules already in the array.
- The interval is visible through `wp_get_schedules()` — the filter is registered at load.
- `wp_presence_schedule_cleanup()` schedules `wp_delete_expired_presence_data` when nothing is scheduled.
- Calling it twice does not double-schedule (the guard on line 16).

Tests are isolated via `wp_clear_scheduled_hook()` in `set_up()`/`tear_down()`, since bootstrap may already have scheduled the event.

Fixes #220

### Testing
Tests only — no production code changed. I don't have a local `wp-env` PHP/MySQL environment, so I couldn't run PHPUnit/PHPCS locally; CI is the authoritative check.

### Use of AI Tools
AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 4.8
Used for: Writing the test file following the suggested direction in #220. I have reviewed it and take responsibility for it.

### DIFF
--- a/tests/test-cron.php
+++ b/tests/test-cron.php
@@ -77,7 +77,7 @@ class WP_Test_Presence_Cron extends WP_UnitTestCase {
 	}
 
 	/**
-	 * The guard on line 16 of includes/cron.php must prevent a second event.
+	 * The guard in wp_presence_schedule_cleanup() must prevent a second event.
 	 *
 	 * @covers ::wp_presence_schedule_cleanup
 	 */
--- a/tests/test-cron.php
+++ b/tests/test-cron.php
@@ -74,6 +74,7 @@ class WP_Test_Presence_Cron extends WP_UnitTestCase {
 		wp_presence_schedule_cleanup();
 
 		$this->assertNotFalse( wp_next_scheduled( self::HOOK ), 'The cleanup event should be scheduled.' );
+		$this->assertSame( 'wp_presence_every_minute', wp_get_schedule( self::HOOK ), 'Cleanup should run on the one-minute interval.' );
 	}
 
 	/**
--- a/tests/test-cron.php
+++ b/tests/test-cron.php
@@ -1,0 +1,115 @@
+<?php
+/**
+ * Tests for cron scheduling (includes/cron.php).
+ *
+ * @package Presence_API
+ *
+ * @group presence
+ * @group cron
+ */
+class WP_Test_Presence_Cron extends WP_UnitTestCase {
+
+	const HOOK = 'wp_delete_expired_presence_data';
+
+	public function set_up() {
+		parent::set_up();
+		// Bootstrap and other tests may have scheduled the cleanup event; start clean.
+		wp_clear_scheduled_hook( self::HOOK );
+	}
+
+	public function tear_down() {
+		wp_clear_scheduled_hook( self::HOOK );
+		parent::tear_down();
+	}
+
+	/**
+	 * @covers ::wp_presence_cron_schedules
+	 */
+	public function test_cron_schedules_adds_every_minute_interval() {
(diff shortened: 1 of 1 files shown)


tests/test-cron.php
```diff
@@ -0,0 +1,115 @@
...
+
+	/**
+	 * @covers ::wp_presence_schedule_cleanup
+	 */
+	public function test_schedule_cleanup_schedules_the_event() {
+		$this->assertFalse( wp_next_scheduled( self::HOOK ), 'No event should be scheduled to begin with.' );
+
+		wp_presence_schedule_cleanup();
+
+		$this->assertNotFalse( wp_next_scheduled( self::HOOK ), 'The cleanup event should be scheduled.' );
```
josephfusco, line 76: Nothing asserts which recurrence the event uses. Changing `'wp_presence_every_minute'` to `'hourly'` in `includes/cron.php` leaves all five tests green, and the TTL is 60 seconds.

```suggestion
		$this->assertNotFalse( wp_next_scheduled( self::HOOK ), 'The cleanup event should be scheduled.' );
		$this->assertSame( 'wp_presence_every_minute', wp_get_schedule( self::HOOK ), 'Cleanup should run on the one-minute interval.' );
```

tests/test-cron.php
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	/**
+	 * The guard on line 16 of includes/cron.php must prevent a second event.
```
josephfusco: Line references go stale.

```suggestion
	 * The guard in wp_presence_schedule_cleanup() must prevent a second event.
```

josephfusco (commented):
Ran it locally against current `main`: 133 tests / 282 assertions single site, 133 / 295 multisite, phpcs clean, `includes/cron.php` at 7/7 statements. Closes #220.

One gap and one nit inline.

Local setup if you want it later: `npm ci && npx wp-env start`, then `npx wp-env run tests-cli --env-cwd='wp-content/plugins/presence-api' -- ./vendor/bin/phpunit`.

josephfusco (approved):
Thank you for these additions!

PR #250, opened 2026-08-09
feat: seed the Playground demos with real repository contributors

josephfusco:
The Playground demos seed themselves with generated names, which seemed like a missed chance to credit the people who show up. Opening as a draft to gauge whether this is wanted at all.

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5
Used for: Assisting with implementation, tests, and verification; reviewed and edited by me.


josephfusco:
Rewritten in 8448d24 to read the contributor list from the `Contributors` line in `readme.txt`. That drops the generator script, its tests, the opt-out file, the `release-please.yml` step, and the runtime GitHub API call. Net diff is now 311 lines across 2 files, down from 856 across 6.

Four things worth a second opinion.

**The seeder writes an mu-plugin.** `wp_presence_demo_install_avatar_filter()` drops `presence-demo-avatars.php` into `WPMU_PLUGIN_DIR` to serve avatars through `pre_get_avatar_data`, and `wp_presence_demo_cleanup()` removes it. The alternative was editing four blueprints plus `playground-preview-publish.yml`. A seeder writing a file is the least conventional part of the diff.

**Avatars depend on `wordpress.org/grav-redirect.php`.** It resolves a WordPress.org username to that person's Gravatar without exposing the address, and passes `&s=` through, verified at 24, 32, and 96px. I have not found it documented as a stable public endpoint, so it may be relying on an implementation detail.

**Contributors appear at the next release, not at merge.** `readme.txt` is updated by the existing `sync-contributors.js` flow on release PRs. Same cadence as props, so nobody waits longer for the demo than for their credit, but it does mean a merged contributor is not in the demo that day.

**People show as their WordPress.org username, not their display name.** `readme.txt` carries usernames only, and resolving them to names would need a request per person on every boot.

Not yet run: `tests/e2e/presence-screenshots.test.js`.

PR #251, opened 2026-08-09
test: cover post list and user list admin integrations

### LINKED ISSUE #219: Post list and user list admin integrations have no tests (0% coverage)
Two admin list table integrations have no test file at all:

- `includes/post-list.php`, 0%, 58 lines missed. Registers an "Editors" column on every public post type list table and renders current editors into it.
- `includes/user-list.php`, 0%, 32 lines missed. Adds an "Online" view link to the users list table and filters `WP_User_Query` when that view is active.

Both are plain function-based hook targets with no `WP_CLI` or `WP_DEBUG` gate, so unlike #218 there is nothing structural stopping them being tested today, and both are directly callable.

**Suggested direction**

`tests/test-post-list.php`:

- `wp_presence_register_post_list_columns()` registers the column filter for public post types and skips non-public ones
- `wp_presence_add_editors_column()` inserts the column in the expected position and preserves the existing columns
- `wp_presence_render_editors_column()` with no presence, with one editor, and with several; assert output is escaped
- The function ignores column names other than its own
- `wp_presence_editors_column_css()` enqueues only on the right `$hook_suffix`

`tests/test-user-list.php`:

- `wp_presence_users_views()` adds the Online view and reports an accurate count
- The view is marked current when the request asks for it
- `wp_presence_filter_online_users()` restricts the query to users with presence, and leaves the query untouched otherwise
- Behaviour when nobody is online

`tests/test-admin-bar.php` is the closest existing model. Its tests carry no `@covers` at all, which is why `includes/admin-bar.php` reports a truthful 71.96%; see #217 before copying annotation style from elsewhere.

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

i-am-chitti:
### Summary

Two admin list-table integrations had 0% coverage — no test file existed for either.

- `includes/post-list.php`: the "Editors" column on public, presence-supporting post types (avatars of who's currently editing).
- `includes/user-list.php`: the "Online" view/filter on the Users list table.

Two things worth calling out for review:
- `wp_presence_render_editors_column()` caches into a function-local `static`, which is never reset between test methods in the same process. All render scenarios (none/one/several editors, escaping, column-name guard) are exercised in a single test method to avoid a stale-cache false pass.
- The online count/filter always includes the current user, even with no presence row of their own (existing behavior in `user-list.php:30-32` and `:78-80`) — tests assert this explicitly rather than treating it as a bug.

Fixes #219

### Testing
- `npm run test`
- `phpcs`

### Use of AI Tools
AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Sonnet 5
Used for: Implementing both test files and PR description


i-am-chitti:
The PHP 8.3 job failed during npx wp-env start, before PHPUnit ran — a transient Alpine package fetch failure while building the tests-cli image. Unrelated to this PR, so a re-run should hopefully sort it.

Separately: since this could hit any PR, happy to look into it once it's confirmed transient.

josephfusco:
Agreed, transient — the re-run from the  commit is already in flight.

josephfusco (approved):
🙌🏼

PR #256, opened 2026-08-10
test: cover the WP-CLI presence command

### LINKED ISSUE #218: WP-CLI command has no tests (0% coverage)
`includes/cli/class-wp-presence-cli-command.php` is at 0% coverage, 61 lines missed. There is no test file for it.

Unlike the widget and admin files, this is a genuine absence rather than an artefact of `@covers` narrowing (see #217). The class is loaded only when `WP_CLI` is defined:

https://github.com/WordPress/presence-api/blob/main/presence-api.php#L97-L100

so it never loads during a PHPUnit run at all.

**Suggested direction**

Two parts, and the first is the real work:

1. Make the class loadable in tests. The `require_once` is inside the `WP_CLI` guard, so a test file needs to require it directly and stub the `WP_CLI` class methods the command calls (`WP_CLI::success`, `WP_CLI::error`, `WP_CLI::line`, and any formatter use). The wp-cli testing convention is to assert on the arguments passed to those static calls.

2. Cover each subcommand: the happy path, the empty result path, and the argument validation path.

Note that #210 tracks a related defect in this file: the CLI queries the presence table with no availability check. Whoever writes these tests should coordinate, since a test asserting graceful behaviour on a missing table is exactly what #210 needs.

Reasonable candidate for a first contribution once the loading question is settled, though the stubbing is fiddly enough that it is not a Good First Issue as it stands.

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

i-am-chitti:
### Summary

This PR adds unit tests across all four WP CLI subcommands — `set`, `list`, `summary`, `cleanup`.

Implemented as the issue suggests: `tests/stubs/wp-cli.php` provides minimal WP-CLI stand-ins, and the test file requires the command class directly rather than defining the `WP_CLI` constant, which `wp_presence_is_admin_screen_save()` branches on.

**No production code touched.**

Fixes #218

### Screenshots

image

image

### Testing
- `npm run test`, and again with `WP_MULTISITE=1`
- `phpunit --coverage-text`
- `phpcs`, `phpstan`

### Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5
Used for: Assisting with implementation


i-am-chitti:
The two failing checks -`phpunit / PHP 7.4` and `e2e / Playwright`- look like a transient runner network issue.

i-am-chitti:
These tests call the subcommand methods directly, so WP-CLI's dispatcher is never exercised - `--format` validation against the `## OPTIONS` synopsis, the `@subcommand list` mapping, declining the `cleanup` prompt, and the command registration in `presence-api.php:97-100`.

Covering that needs Behat functional tests via [`wp-cli/wp-cli-tests`](https://github.com/wp-cli/wp-cli-tests). It's scoped to WP-CLI command packages rather than plugins, so it'd mean a dev dependency, a separate test database, and another CI job.

Worth a follow-up issue, or is that layer WP-CLI's to test rather than ours?

josephfusco:
`--format` validation and the `@subcommand list` mapping are WP-CLI parsing your docblocks, not code this plugin owns. That's wp-cli/wp-cli-tests' Behat suite's job, not a second CI harness here.

Command registration is one `WP_CLI::add_command()` call behind the `WP_CLI` guard. The stub deliberately doesn't define that constant because it changes `wp_presence_is_admin_screen_save()`'s behavior for every other test in the suite. Testing that line would mean breaking that isolation to confirm a single function call happened.

The `confirm()` prompt is presence-api's own behavior, not WP-CLI's. `cleanup()` calls `WP_CLI::confirm()` and runs the DELETE immediately after, no branching, relying on the real `confirm()` to halt on "no." The stub's `confirm()` just records and returns, so there's no way to test "user declines, nothing gets deleted." Fixable without Behat: the stub's `confirm()` can throw the same halt exception `error()` already uses, then a test can assert the row count is unchanged. Opened #260.

josephfusco (approved):
Went through this locally too. The design call to require the command class directly instead of defining WP_CLI is the right one, I checked and that constant really does change behavior for wp_presence_is_admin_screen_save() elsewhere, so defining it globally would've been a much messier trade. Traced a few of the trickier assertions against the actual source (the write-failure test's db_version filter trick, the empty-array-to-'[]' JSON round trip in list) and they all hold up.

Ran phpcs, phpstan, the full suite, and a multisite pass locally, all clean, and the CLI command file lands at 100% coverage now.

i-am-chitti:
Gotcha! Agree `cleanup()` is our behaviour rather than WP-CLI's, since the DELETE runs unconditionally after `confirm()`. I can pick #260 once this lands.

josephfusco (approved):

Issue #203, opened 2026-08-08
Front-end and login writes hit the presence table before it is created


josephfusco:
Split out of #137, which tracked this alongside open-ended audit questions. This part is a confirmed, reproducible defect rather than something needing investigation, so it is broken out to be actionable on its own.

## The problem

Table creation is hooked only to admin contexts:

```php
add_action( 'admin_init', 'wp_maybe_create_presence_table' );  // presence-api.php:170
add_action( 'cli_init', 'wp_maybe_create_presence_table' );    // presence-api.php:171
```

`register_activation_hook()` also calls it, but on network activation that fires once in the network admin and seeds only the main site. `wp_presence_db_version` is per-site, so every other site waits on its own `admin_init`.

Two write paths never pass through `admin_init`:

- `wp_presence_enqueue_heartbeat_ping()` is registered on **both** `admin_enqueue_scripts` and `wp_enqueue_scripts` (`presence-api.php:179-180`) and writes at `includes/heartbeat.php:103`
- `wp_presence_on_login()` writes at `includes/lifecycle.php:23`

On a network-activated site where no admin has loaded wp-admin, a logged-in front-end view or a login writes to a missing table and surfaces a `wpdb` error.

## Reproduction

1. Multisite network, plugin network-activated
2. Create a new site and do not visit its wp-admin
3. Log in and view the front end as a logged-in user on that site

## Notes

Verified against `main` at 0.1.14. The remaining items in #137, cross-site capability behavior and confirming site-scoping with a test, stay there.

PR #257, opened 2026-08-10
fix: guard CLI cleanup and DB viewer with wp_presence_has_table()

### LINKED ISSUE #210: CLI and debug viewer query the presence table without an availability check
The WP-CLI command and the debug DB viewer query the presence table directly with no availability check:

- `includes/cli/class-wp-presence-cli-command.php:210`
- `includes/db-viewer.php:35` and `includes/db-viewer.php:39`

In practice `cli_init` and `admin_init` provision the table before either can run, so this is not a live bug. It is an inconsistency: every other read and write path checks `wp_presence_has_table()` first and degrades to an empty result, so these two are the only places left that would surface a raw `wpdb` error if the table were missing.

Deliberately left out of the fix for #203 to keep that change scoped.

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

aldorizona10-glitch:
## What & why

Resolves #210. The WP-CLI `cleanup` command (`includes/cli/class-wp-presence-cli-command.php`) and the debug DB viewer (`includes/db-viewer.php`) were the only two paths that hit the presence table directly without a `wp_presence_has_table()` check. Every other read/write path in `functions.php` and the REST controller guards on it and degrades to an empty result, so these two were the last places that would surface a raw `wpdb` error if the table were ever missing.

As #210 notes, `cli_init` / `admin_init` provision the table before either can run, so this is a **consistency fix**, not a live bug — it was deliberately left out of #203 to keep that change scoped.

## Changes

- **`db-viewer.php`** — wrap the `SELECT` / `COUNT(*)` queries in the availability check; when the table is absent, fall back to an empty row set and a zero count. The template already renders an empty state (`No entries.`) and the timestamp `foreach` is `empty()`-guarded, so no markup changes are needed.
- **`cli` `cleanup()`** — short-circuit to a `0 entries deleted.` success before the destructive `DELETE` (and before the confirmation prompt, since there is nothing to delete).

No new strings, no schema changes, no behavior change on a provisioned site.

## Testing

WP-CLI is not loaded in the PHPUnit suite and CLI coverage is being handled separately (#218 / #256), so I've kept this change focused on the guards rather than adding CLI-invoking tests here. Happy to add coverage if you'd prefer a particular approach.

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 4.8
Used for: Investigating the guard pattern and drafting the change

### DIFF
--- a/includes/cli/class-wp-presence-cli-command.php
+++ b/includes/cli/class-wp-presence-cli-command.php
@@ -202,6 +202,12 @@ class WP_Presence_CLI_Command extends WP_CLI_Command {
 	public function cleanup( $args, $assoc_args ) {
 		global $wpdb;
 
+		if ( ! wp_presence_has_table() ) {
+			/* translators: %d: Number of deleted entries. */
+			WP_CLI::success( sprintf( __( '%d entries deleted.', 'presence-api' ), 0 ) );
+			return;
+		}
+
 		if ( ! WP_CLI\Utils\get_flag_value( $assoc_args, 'yes', false ) ) {
 			WP_CLI::confirm( __( 'This will delete all presence data. Continue?', 'presence-api' ) );
 		}
--- a/includes/db-viewer.php
+++ b/includes/db-viewer.php
@@ -29,14 +29,22 @@ add_action(
 
 		global $wpdb;
 
-		// No user input in these queries; table name comes from $wpdb->presence (controlled).
-	// phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
-		$rows = $wpdb->get_results(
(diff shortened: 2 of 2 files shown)


aldorizona10-glitch:
@i-am-chitti you're right, and apologies for the duplication. I went straight to a PR without commenting on #210 to claim it first, which is exactly what CONTRIBUTING asks — that's on me. Since you called it first and already had it written, I'll close this in favor of your PR. I'll comment to claim before starting next time. Thanks for the nudge.

i-am-chitti:
Thanks @aldorizona10-glitch for the graceful close. We'd independently landed on almost the same guard, which was oddly reassuring.
Opening mine shortly. Happy for you to be in props on it since you did the same work.

PR #258, opened 2026-08-10
fix: delete expired presence rows by key in bounded passes

### LINKED ISSUE #206: Expired-data cleanup uses MySQL-only DELETE ... LIMIT in an uncapped loop
`wp_delete_expired_presence_data()` (`includes/functions.php`) deletes in batches:

```php
do {
	$deleted = $wpdb->query(
		$wpdb->prepare( "DELETE FROM {$wpdb->presence} WHERE date_gmt < %s LIMIT 1000", $cutoff )
	);
} while ( $deleted > 0 );
```

Two divergences from core:

1. **`DELETE ... LIMIT` does not appear anywhere in `wp-includes`.** It is MySQL-specific and not standard SQL, so it is a portability hazard for any non-MySQL backend, including the SQLite integration that Playground uses to run the demo blueprints.
2. **The loop has no iteration ceiling.** A site returning from a cron outage with a large backlog will keep looping until the request hits `max_execution_time`, having committed an unpredictable fraction of the work.

Core's batching convention is to select a bounded set of primary keys and then delete by key, with a fixed number of passes per invocation and the remainder left for the next cron run.

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

aldorizona10-glitch:
## What & why

Resolves #206. `wp_delete_expired_presence_data()` (`includes/functions.php`) batched deletions with a MySQL-only `DELETE ... LIMIT` inside a loop that had no iteration ceiling:

```php
do {
	$deleted = $wpdb->query(
		$wpdb->prepare( "DELETE FROM {$wpdb->presence} WHERE date_gmt < %s LIMIT 1000", $cutoff )
	);
} while ( $deleted > 0 );
```

Two divergences from core, both from #206:

1. **Portability** — `DELETE ... LIMIT` is MySQL-specific and appears nowhere in `wp-includes`, so it breaks on non-MySQL backends, including the SQLite integration Playground uses to run the demo blueprints.
2. **Unbounded work** — a site returning from a cron outage with a large backlog keeps looping until `max_execution_time`, committing an unpredictable fraction of the work.

## Change

Follows core's batching convention — *select a bounded set of primary keys, delete by key, a fixed number of passes per invocation, remainder to the next run*:

- Each pass `SELECT`s a page of `id`s older than the cutoff (`LIMIT %d`) and deletes them with a portable `WHERE id IN ( ... )` built from `%d` placeholders passed through `prepare()`.
- A fixed **max passes per invocation** bounds the work; the cron event fires every minute, so any remaining backlog drains over subsequent runs instead of in one long request.
- Batch size (default `1000`) and passes per run (default `10`) are filterable via `wp_presence_cleanup_batch_size` and `wp_presence_cleanup_max_passes`.

No schema change; behavior on a normally-drained table is unchanged.

## Tests

Added to `tests/test-functions.php` (`@covers ::wp_delete_expired_presence_data`):

- **`test_cleanup_is_bounded_per_invocation`** — with `batch_size` and `max_passes` filtered to `2`, seeds 5 expired rows and asserts one invocation deletes at most `2 × 2 = 4`, leaving the remainder; a second invocation clears it. This locks in the "remainder to the next run" guarantee.
- **`test_cleanup_leaves_fresh_entries_untouched`** — fresh rows survive cleanup even with a batch size of one.

The existing `test_cleanup_removes_expired_entries` continues to pass.

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 4.8
Used for: Investigating core's batching convention, implementing the fix, and writing tests

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -421,6 +421,15 @@ function wp_get_presence_summary( $timeout = WP_PRESENCE_DEFAULT_TTL ) {
 /**
  * Deletes stale presence entries older than the default TTL.
  *
+ * Runs on the every-minute cron event. Rather than looping without a ceiling
+ * over the MySQL-only `DELETE ... LIMIT` construct, this selects a bounded
+ * page of primary keys older than the cutoff and deletes them by key, for a
+ * fixed number of passes per invocation. Any remaining backlog is left for the
+ * next cron run, so a single request cannot run until `max_execution_time`
+ * when a site returns from a cron outage with a large backlog. Deleting by
+ * primary key also keeps the query portable to non-MySQL backends, such as the
+ * SQLite integration Playground uses to run the demo blueprints.
+ *
  * @access private
  */
 function wp_delete_expired_presence_data() {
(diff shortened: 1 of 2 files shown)


josephfusco (approved):
Nice fix. Traced the bounded loop by hand against the 5-row test case and it checks out exactly (4 deleted then 1), and the dynamic IN() list is safe since only ints ever get interpolated into the placeholder skeleton, the values go through prepare() properly.

Ran phpcs, phpstan, and the full suite locally too, all clean. Approving.

PR #261, opened 2026-08-11
fix: check table availability in the CLI command and debug viewer

### LINKED ISSUE #210: CLI and debug viewer query the presence table without an availability check

i-am-chitti:
Adds the `wp_presence_has_table()` check to the last two paths that queried the presence table without one - the WP-CLI `cleanup` command and the debug DB viewer. Both now degrade to an empty result like every other read path.

The viewer's `SELECT COUNT(*)` is removed rather than guarded as `$count` was never read - https://github.com/WordPress/presence-api/blob/1f3de12d576ec469e700ebd6c1bb522f18d58489/includes/db-viewer.php#L38-L39

No test for the CLI `tests/test-cli-command.php` yet. I'll add once WP #256 is merged.

Fixes #210

## Use of AI Tools

AI assistance: No

### DIFF
--- a/includes/cli/class-wp-presence-cli-command.php
+++ b/includes/cli/class-wp-presence-cli-command.php
@@ -202,6 +202,12 @@ class WP_Presence_CLI_Command extends WP_CLI_Command {
 	public function cleanup( $args, $assoc_args ) {
 		global $wpdb;
 
+		if ( ! wp_presence_has_table() ) {
+			/* translators: %d: Number of deleted entries. */
+			WP_CLI::success( sprintf( __( '%d entries deleted.', 'presence-api' ), 0 ) );
+			return;
+		}
+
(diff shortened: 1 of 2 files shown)


josephfusco (approved):
Nice additions!

PR #263, opened 2026-08-11
test: cover cleanup without presence table

### LINKED ISSUE #262: Add test coverage for the no-table branch in WP_Presence_CLI_Command::cleanup()
#261 adds an early return in `cleanup()` for when the presence table doesn't exist, but doesn't add a test for it — codecov shows 0% patch coverage on those 3 lines as a result.

#256 already established the pattern for simulating this: `add_filter('option_wp_presence_db_version', '__return_zero')`, then assert on `WP_CLI::messages(...)`. Something like:

```php
public function test_cleanup_reports_zero_without_a_table() {
    add_filter( 'option_wp_presence_db_version', '__return_zero' );
    $this->command->cleanup( array(), array( 'yes' => true ) );
    $this->assertSame( array( '0 entries deleted.' ), WP_CLI::messages( 'success' ) );
    $this->assertSame( array(), WP_CLI::messages( 'confirm' ) );
}
```

Follow-up from #261.

floze-the-genius:
Adds regression coverage for the `cleanup()` early return when the presence table is unavailable. The test uses the established database-version filter pattern and verifies both the zero-deletion success message and that no confirmation prompt is shown.

Fixes #262

## Testing

- focused PHPUnit: 1 test, 2 assertions
- full single-site PHPUnit: 187 tests, 417 assertions, 5 expected multisite skips
- PHPCS: 17/17 files clean
- PHPStan: no errors
- workflow script tests: 54/54 passed
- local Playwright E2E: 13/13 passed
- `git diff --check`

The new cleanup case also passes in multisite mode. The complete local multisite run has one unrelated failure in `test_new_site_is_provisioned_on_initialize_site`; it reproduces when run alone, and both that test and its production path are unchanged from `main`. The authoritative hosted PHP 8.3 multisite job passes.

The hosted Playwright job did not reach E2E tests because `wp-env start` encountered a runner/network TLS hostname mismatch while Composer fetched a dependency. The equivalent local Playwright suite passes all 13 tests.

## Use of AI Tools

AI assistance: Yes  
Tool(s): OpenAI Codex  
Model(s): GPT-5  
Used for: Drafting the issue-prescribed regression test and validation notes, running the relevant project gates, and an additional AI-assisted review. Human maintainer review remains required.


josephfusco:
Hey @floze-the-genius thanks for tackling this.

>The following contributors have not linked their GitHub and WordPress.org accounts: @floze-the-genius.

To receive proper attribution please associate a WordPress.org account with your GitHub account. Let me know once that's setup and I'll run the props-bot again.

tests/test-cli-command.php
```diff
@@ -302,6 +302,18 @@ public function test_cleanup_deletes_every_entry_regardless_of_age() {
 		$this->assertSame( array( '3 entries deleted.' ), WP_CLI::messages( 'success' ) );
 	}
 
+	/**
+	 * @covers WP_Presence_CLI_Command::cleanup
+	 */
+	public function test_cleanup_reports_zero_without_a_table() {
+		add_filter( 'option_wp_presence_db_version', '__return_zero' );
+
+		$this->command->cleanup( array(), array( 'yes' => true ) );
```
i-am-chitti: The confirm assertion here doesn't actually test the no-table code path. 

`cleanup()` skips `confirm()` for two reasons - either the table is missing, or `'yes' => true` was passed. This test passes `'yes' => true`, so `confirm()` is already skipped, regardless of whether the no-table path fires. I confirmed this by removing the no-table early return from `cleanup()` and rerunning. It still passed.

The "yes flag skips confirm" behavior is already covered here  [test_cleanup_skips_the_prompt_with_the_yes_flag](https://github.com/WordPress/presence-api/blob/8310ba724d1af7d99c8d8f5ca3f648cbd3959a38/tests/test-cli-command.php#L308-L312) .


```suggestion
		$this->command->cleanup( array(), array() );
```

josephfusco (approved):
Traced the filter through to `wp_presence_has_table()` → `(bool) get_option('wp_presence_db_version')`, confirms it actually exercises the early-return path. Matches the existing filter pattern in this file. Small, correct, green.

josephfusco:
Merged. Link your WordPress.org account any time before 0.2.0 ships and comment here, and re-running props-bot will pick you up for the release props.

PR #271, opened 2026-08-12
fix: preserve focus across heartbeat re-renders in dashboard widgets

### LINKED ISSUE #265: Heartbeat swap drops focus to <body> in dashboard widgets
Live-tested with Playwright (real focus + heartbeat tick, no code-only inference). Part of the #138 audit.

Both dashboard widgets (Active Posts, Who's Online) replace their entire list markup on every heartbeat tick where the data changes, via `container.html(...)`. Nothing captures or restores focus around that replacement.

**Before the tick** — focus is on the post-title link:

focus before tick

**After a data change + heartbeat tick** — the list re-rendered (now shows "2 editors" instead of "admin"), and focus has silently fallen back to `<body>`:

focus lost to body

Anyone tabbing through the dashboard who happens to have focus inside either widget when a tick lands (heartbeat runs roughly every 15s while active) gets dropped back to the top of the document with no warning.

**Cause**: `class-wp-presence-widget-active-posts.php` and `class-wp-presence-widget-whos-online.php`'s inline scripts call `container.html(buildFullPostsHtml(...))` / `container.html(buildFullHtml(...))` unconditionally on signature change, full innerHTML replacement.

**Fix direction**: capture `document.activeElement` before the swap; if it was inside the container, restore focus to the equivalent item after rebuild (matched by id/data attribute) or fall back to the container itself rather than `<body>`.

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Sonnet 5
Used for: Assisting with investigation, live browser verification, and drafting

i-am-chitti:
## Description

- Capture focus before `container.html()` swaps in Active Posts and Who's Online, restore it after - falling back to the container if the focused item no longer exists.
- Add `data-post-id` to Active Posts rows (Who's Online already had `data-user-id`); both containers get `tabindex="-1"` as the fallback focus target.
- Add e2e coverage for focus preservation across a heartbeat re-render in both widgets.

Fixes #265.

## Test plan
- `npx playwright test --config tests/e2e/playwright.config.js tests/e2e/presence-widgets.test.js`

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5
Used for: Investigating the focus-loss cause and writing the e2e tests


josephfusco (approved):
Nice work!

PR #275, opened 2026-08-12
fix: keyboard users can't reach non-link items in the admin bar presence flyout

### LINKED ISSUE #264: Admin bar flyout: non-link items unreachable by keyboard
Live-tested with Playwright (headless Chromium, keyboard events, no code-only inference). Part of the #138 audit.

The flyout opens correctly via Enter/Space on the toolbar node. Once open, Tab skips every non-link item (self, "On this page"/"Elsewhere" headers, users without a discoverable screen) and jumps straight to the first real `<a>` in the flyout.

**Before Tab** (flyout open, focus on the toolbar node itself):

flyout open

**After one Tab press** — focus jumps past "On this page," "admin (you)," "Elsewhere," and "User B" straight to "View online users":

tab skips to bottom link

A keyboard-only user has no way to read who is online or on which screen; they can only ever reach the one link at the bottom.

**Cause**: `includes/admin-bar.php` builds child nodes with `href => false` for anything without a real destination (self, per-user "here" entries, group headers). WP_Admin_Bar renders those as `<div class="ab-item ab-empty-item">` (`wp-includes/class-wp-admin-bar.php`), which is never in the tab order, unlike `<a href>`.

**Fix direction**: give informational-only flyout items a `tabindex="-1"` (or similar) plus a way for a keyboard user to actually read them in order — e.g. `role="none"` text nodes read via a wrapping `aria-label`/`aria-describedby` on the parent, or restructure so Tab moves through every item in visual order regardless of whether it links anywhere.

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Sonnet 5
Used for: Assisting with investigation, live browser verification, and drafting

i-am-chitti:
## Description

Fixes #264

Non-link flyout items (group headers, self row, per-user rows without a screen link) had no `tabindex`, so Tab skipped them and jumped straight to the last real link.

- Add `meta.tabindex => 0` to every non-link node.
- Adjust the group-header CSS so focus color is visible instead of being overridden.
- Add a PHPUnit test asserting every href-less node stays keyboard reachable.

## Screencast

https://github.com/user-attachments/assets/9c8b0cd3-7ec6-41b3-b588-3dea051d0d14

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Sonnet 5
Used for: Investigating the root cause and drafting the fix.

### DIFF
--- a/includes/admin-bar.php
+++ b/includes/admin-bar.php
@@ -317,8 +317,6 @@ function wp_presence_admin_bar_node( $wp_admin_bar ) {
 					'id'     => 'presence-user-' . $entry->user_id,
 					'title'  => $item_title,
 					'href'   => $screen_url ? $screen_url : false,
-					// Non-link items need an explicit tabindex; WP_Admin_Bar only
-					// renders one when it's set, otherwise the item is skipped by Tab.
 					'meta'   => $screen_url ? array() : array( 'tabindex' => 0 ),
 				)
 			);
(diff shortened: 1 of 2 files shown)


josephfusco:
I rebased this to clear the the workflow errors from the 7.0.4 blip.

josephfusco (approved):
🙌🏼

Issue #205, opened 2026-08-08
Presence aggregates run unbounded and uncached on every heartbeat tick


josephfusco:
`wp_get_presence_summary()` and `wp_get_active_rooms()` in `includes/functions.php` both select every unexpired row and aggregate in PHP. Neither applies a `LIMIT`, and neither caches.

The REST collection endpoint paginates, but the dashboard widgets go through these two functions, so each heartbeat tick on an open dashboard scans the entire live presence set.

The plugin's documented decision not to cache presence reads is sound for `wp_get_presence()`, where the data is ephemeral and a stale answer is worse than a query. It does not extend to an unbounded aggregate whose cost grows with concurrent users, which is exactly the metric this feature scales on.

Core's pattern for an expensive aggregate is a bounded query plus a short-lived cache entry.

- [x] `wp_get_presence_summary()` — aggregated in SQL with `GROUP BY` instead of pulling every row into PHP. Fixed by #277.
- [x] `wp_get_active_rooms()` — same unbounded shape, same fix pattern. `get_rooms()`'s hydrate-then-slice pagination is a related but separate problem, split to #285.

Related to #132, which will need numbers on this.

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5
Used for: Assisting with investigation and drafting

i-am-chitti:
Hi @josephfusco, I would like to pick this up. Looks interesting to work on.

My leaning: GROUP BY as the actual fix, capping only for the member lists in wp_get_active_rooms() (an avatar per user per room can't be aggregated away), and hold the cache until #132 has numbers. GROUP BY room + COUNT(DISTINCT user_id) is portable, so the no-SUBSTRING_INDEX() constraint survives. One callout: distinct users don't sum across rooms, so exact per-prefix counts need a small query per prefix.

Two things:

- get_rooms() hydrates every room before it array_slice()s [class-wp-rest-presence-controller.php](https://github.com/WordPress/presence-api/blob/cfd705bc66e352c4b56e301500848edd87eec08d/includes/class-wp-rest-presence-controller.php#L653), so it's unbounded too, despite the docs calling it "SQL pagination".
- A non-persistent group is request-local, so it won't survive between heartbeat ticks. Only helps as a persistent group, and only on hosts with a drop-in.

So: is the REST path in scope here or a separate issue? It's the bigger half of the work.

i-am-chitti:
[...]

PR #277, opened 2026-08-13
perf: aggregate wp_get_presence_summary() in SQL

i-am-chitti:
Related: #205 

Rewrites `wp_get_presence_summary()` to aggregate in SQL instead of pulling every unexpired row into PHP. Query count now scales with room/prefix count, not concurrent users.

- `GROUP BY room` for entry counts instead of one row per entry
- `COUNT(DISTINCT user_id)` for the site-wide total
- one `COUNT(DISTINCT user_id) ... room IN (...)` per prefix, since distinct users don't sum across rooms
- stays portable: no `SUBSTRING_INDEX()`, prefix grouping stays PHP-side

Regression test added asserting query count doesn't grow with entry count.

Partially addresses #205. `wp_get_active_rooms()` and the REST rooms endpoint have the same shape and need a separate pass. Following up on the issue.

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Sonnet 5
Used for: Investigation, implementation, and tests

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -374,47 +374,55 @@ function wp_get_presence_summary( $timeout = WP_PRESENCE_DEFAULT_TTL ) {
 	$cutoff  = gmdate( 'Y-m-d H:i:s', time() - $timeout );
 
 	// phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
-	$rows = $wpdb->get_results(
+	$room_rows = $wpdb->get_results(
 		$wpdb->prepare(
-			"SELECT room, user_id FROM {$wpdb->presence} WHERE date_gmt > %s ORDER BY room ASC",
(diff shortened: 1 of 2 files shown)


josephfusco:
Going to release 0.1.19 next, and save this for the following release

josephfusco:
Pushed a fix for the regression test. The original assertion compared num_queries before and after adding entries to existing rooms, but the pre-fix version already issued exactly one query regardless of entry count, so that comparison passes against either implementation and doesn't actually catch a revert. I checked this by reverting just includes/functions.php locally and rerunning the test.

Replaced it with a check on query shape instead: capture the SQL via the query filter and assert every query touching the presence table uses GROUP BY or COUNT(), which is the actual thing that changed. Confirmed it fails against the old implementation and passes against this one.

Everything else here looks solid, the cross-room distinct-user dedup case (the trickiest part of this rewrite) is already covered by the existing test_get_presence_summary test and passes.

i-am-chitti:
Thanks @josephfusco for catching the query-count assertion and fixing it. Verified locally against both implementations and it correctly fails on the old code, passes on the new. Appreciate it 🙌 !

Issue #267, opened 2026-08-12
Reduce Heartbeat polling cost on cheap hosting


josephfusco:
Presence polling pays a full query (and, if riding Heartbeat, part of the WP bootstrap) on every tick, even though most ticks in a quiet room find nothing changed. On cheap shared hosting (few PHP-FPM workers, no guaranteed object cache), that cost matters more than on managed hosting, and it compounds if other features end up polling on a separate schedule instead of sharing this one.

Three independent techniques, tracked as sub-issues below. Each needs its own before/after measurement (ties to #132) rather than shipping on intuition, since they're not equally impactful.

- [x] Skip the response body when presence state hasn't changed
- [x] Coalesce polling across tabs of the same user
- [x] Back off polling interval for idle rooms

Tracking issue only — no code here.

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Sonnet 5
Used for: Assisting with investigation and drafting

i-am-chitti:
@josephfusco I would like to pick up all sub-issues. I've already started on #268.

josephfusco:
@i-am-chitti Thank you for knocking these out! Huge performance win here beyond Presence API too.  Closing as all subtasks have been completed.

PR #281, opened 2026-08-14
fix: guard the presence table upgrade with a provisioning lock

### LINKED ISSUE #209: dbDelta upgrade has no concurrency guard
`wp_maybe_create_presence_table()` has no guard against concurrent execution: two admin requests during a version bump both pass the check and both run `dbDelta()`.

Core's `wp_upgrade()` has the same unguarded pattern, but its exposure is narrowed by a manual confirmation click on `wp-admin/upgrade.php`, a step `admin_init` doesn't have. No core precedent to copy.

`dbDelta()` is idempotent and this is low-likelihood, so not a proposal blocker. A transient or `wp_cache_add()` lock would close it.

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

jigneshbhavani:
`wp_maybe_create_presence_table()` runs on `admin_init` and `cli_init`. Neither has the manual confirmation step that narrows core's own upgrade routine on `wp-admin/upgrade.php`, so two requests arriving during a version bump could both pass the version check and both run `dbDelta()`.

This takes an exclusive lock before the schema change and releases it after `wp_presence_db_version` is written. The loser of the race returns rather than waiting, and the next request repairs anything left over.

## Why an options row rather than `wp_cache_add()`

`wp_cache_add()` only coordinates across requests on sites running a persistent object cache. Under the default non-persistent cache it is per request, so it would give no protection on exactly the cheap shared hosting #267 is about. The insert relies on the unique index on `option_name` instead, so it holds either way.

## Why inline rather than `WP_Upgrader::create_lock()`

Per the discussion on the issue. The implementation mirrors `WP_Upgrader::create_lock()` (`wp-admin/includes/class-wp-upgrader.php`, since 4.5.0), which core uses for the same purpose in `WP_Core_Upgrader::upgrade()` and `WP_Automatic_Updater::run()`. The second is the closest match, since it runs from cron with no confirmation step at all.

## Abandoned locks

A request that dies between taking the lock and releasing it would otherwise leave the site unprovisionable. The lock is reclaimed once it is older than a minute, the same age based recovery core's lock uses. A minute is short because `dbDelta()` on one table is fast.

## Portability

`INSERT IGNORE` is MySQL syntax, and the plugin cares about the SQLite backend Playground uses for the demo blueprints. The SQLite integration translates `INSERT IGNORE` to `OR IGNORE` and has test coverage for it, so the demos are unaffected.

## Testing

Four tests added to `tests/test-table-creation.php`:

- a request that cannot take the lock does not run `dbDelta()` and does not claim the site is provisioned
- provisioning releases the lock so a later request is not blocked
- a lock older than the timeout is reclaimed rather than respected forever
- the lock is exclusive while held and free after release

The two that are regression guards rather than drivers were checked by mutation: removing the release call fails the second, disabling the age check fails the third.

`set_up()` now clears the lock option. DDL commits the surrounding transaction, so a lock left behind by an earlier run survives into the next one and would block every test that provisions.

Full suite: 190 tests, 424 assertions, 5 skipped (multisite only). PHPCS and PHPStan clean.

Fixes #209

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5
Used for: Investigating core's locking precedent, implementing the lock, and writing the tests. I verified the behavior against a local wp-env install, confirmed each test fails before the change and passes after, mutation checked the two regression guards, and ran the full suite plus PHPCS and PHPStan.

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -510,6 +510,88 @@ function wp_presence_table_exists() {
 }
 
 /**
+ * Takes an exclusive lock, or reports that another request already holds it.
+ *
+ * Mirrors WP_Upgrader::create_lock(), which core uses to keep its own upgrade
+ * routines from running twice over, in WP_Core_Upgrader::upgrade() and in
+ * WP_Automatic_Updater::run(). The insert is the lock: option_name is unique,
+ * so exactly one concurrent caller can create the row. That also means it holds
+ * on sites with no persistent object cache, where wp_cache_add() is per request
+ * and would coordinate nothing.
+ *
+ * Written out here rather than calling WP_Upgrader::create_lock() so that
+ * provisioning does not have to load the whole upgrader stack on admin_init and
+ * cli_init for the sake of one static method.
+ *
+ * @access private
+ *
+ * @global wpdb $wpdb WordPress database abstraction object.
+ *
+ * @param string $lock_name       Name of the lock.
+ * @param int    $release_timeout Optional. Seconds after which an unreleased
+ *                                lock is treated as abandoned. Default
+ *                                MINUTE_IN_SECONDS.
+ * @return bool Whether the lock was taken.
+ */
+function wp_presence_create_lock( $lock_name, $release_timeout = null ) {
+	global $wpdb;
+
+	if ( ! $release_timeout ) {
+		$release_timeout = MINUTE_IN_SECONDS;
+	}
+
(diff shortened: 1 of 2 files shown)


josephfusco (approved):
This looks good to me, thank you for working on this!

Issue #204, opened 2026-08-08
Screen revision counter uses a read-modify-write on a single shared option


josephfusco:
`wp_presence_bump_screen_revision()` (`includes/screen-revisions.php:88-111`) does a read-modify-write on a single option: it loads the entire `wp_presence_screen_revisions` map, mutates one key, and writes the whole map back.

It runs on `post_updated`, `edited_term`, `edit_comment`, `profile_update`, and `updated_option`, so it is on the hot path for essentially every content change on the site.

Two problems:

1. **Lost updates.** There is no locking or atomicity between the `get_option()` and the `update_option()`. Two edits that overlap will drop one of the two revision bumps, and the stale-screen banner that depends on the counter will not fire.
2. **Cache churn.** Every bump invalidates the entire option in the object cache, no matter which screen key changed. A post save invalidates the revision data for every settings page, term, and comment along with it.

This diverges from core. Core does not funnel a per-object counter through one shared option, it stores per-object state on the object.

## Direction

No new table, and no new writes on the hot path.

Moving the map into postmeta is not the fix on its own. `update_metadata()` calls `wp_cache_delete( $object_id, $meta_type . '_meta' )` (`wp-includes/meta.php:145`), so a write invalidates every cached meta value for that object, not just ours. On a post carrying a lot of meta that trades one shared invalidation for a per-post one.

**Posts need no stored revision at all.** Everything the banner reads already exists and is already maintained by core on the same save:

- Revision: `post_modified_gmt`, set on every update (`wp-includes/post.php:4698`).
- Actor: `_edit_last`, written by `edit_post()` (`wp-admin/includes/post.php:447`).
- Time: `post_modified_gmt`.

Both are read through the post and meta caches that the edit screen has already primed, so the heartbeat lookup costs no extra query and the save costs no extra write. `post_updated` is the overwhelming majority of bumps, so this removes most of the churn rather than relocating it.

The remaining four have no equivalent and do need storage, but none of them are hot:

| Key | Storage | Note |
| --- | --- | --- |
| `post/{id}` | none | `post_modified_gmt` + `_edit_last` |
| `user-edit/{id}` | usermeta | `wp_insert_user()` already writes user meta on every profile update, so the bucket is dirty regardless |
| `term/{tax}/{id}` | termmeta | new invalidation, but term edits are rare |
| `comment/{id}` | commentmeta | new invalidation, but comment edits are rare |
[...]

PR #283, opened 2026-08-14
fix: stop funneling screen-revision bumps through one shared option

josephfusco:
## Summary

Fixes a bug where saving any admin screen could silently drop another screen's stale-notice, since every screen's revision lived in one shared, read-modify-write option (#204).

## Validation

- phpcs / phpstan: clean.
- 196/196 tests, single-site and multisite.
- Verified against core: options-page whitelist matches core's own `$allowed_options` gate in `wp-admin/options.php`; post revision reads cost one query cold, not two; timestamp conversion is timezone-safe.

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Sonnet 5
Used for: Implementation, self-review, and verifying design decisions against WordPress core source

### DIFF
--- a/includes/screen-revisions.php
+++ b/includes/screen-revisions.php
@@ -168,10 +168,10 @@ function wp_presence_get_screen_revision( $screen_key ) {
 			);
 
 		case 'user':
-        case 'term':
-        case 'comment':
-       $entry = get_metadata( $target['type'], $target['id'], '_wp_presence_screen_rev', true );
-       return $entry ? $entry : null;
+		case 'term':
+		case 'comment':
+			$entry = get_metadata( $target['type'], $target['id'], '_wp_presence_screen_rev', true );
+			return $entry ? $entry : null;
 
 		case 'options':
 			$entry = get_option( 'wp_presence_screen_rev_options_' . $target['page'], null );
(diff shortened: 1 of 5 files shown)


includes/screen-revisions.php
```diff
@@ -31,32 +31,162 @@
...
+			$entry = get_user_meta( $target['id'], '_wp_presence_screen_rev', true );
+			return $entry ? $entry : null;
+
+		case 'term':
+			$entry = get_term_meta( $target['id'], '_wp_presence_screen_rev', true );
+			return $entry ? $entry : null;
+
+		case 'comment':
+			$entry = get_comment_meta( $target['id'], '_wp_presence_screen_rev', true );
+			return $entry ? $entry : null;
```
i-am-chitti: nit: the user/term/comment here are repeated, differing only in which meta function is called. Since `$target['type']` already holds 'user'|'term'|'comment', these can be combined into one case using core's [get_metadata](https://developer.wordpress.org/reference/functions/get_metadata/)

```suggestion
		case 'user':
        case 'term':
        case 'comment':
       $entry = get_metadata( $target['type'], $target['id'], '_wp_presence_screen_rev', true );
       return $entry ? $entry : null;
```

i-am-chitti (approved):
LGTM 🚀

Issue #131, opened 2026-07-26
Add state-change detection before upsert to reduce write volume


josephfusco:
Presence is upserted unconditionally, with no state comparison. Writes come from two paths, not just heartbeat ticks:

| Trigger | Writes |
|---|---|
| Any admin or front-end render | 1 (`admin/online`, `includes/heartbeat.php:103`) |
| Post-edit screen render | 2 (+ `editor-{id}`, `:108`) |
| Heartbeat tick | 1 (`admin/online`) |
| Heartbeat tick, post-edit screen | 3 (+ `editor-{id}` + `lock-{id}`) |

Most of these change only `date_gmt`. The three-write case is the duplicate entry in #134; resolving that removes one write per tick before anything here is evaluated.

## Options

- Client-side: hash current state, send only on change or near TTL expiry.
- Server-side: skip the write if incoming state matches stored state and `date_gmt` is not near cutoff.
- Accept the volume and document the rationale.

## Constraints

**The ping is also the keepalive.** Any solution gating writes on state change must still write before TTL expiry, or active users appear offline.

**A client-side diff cannot cover render-time writes**, which happen server-side during the page request. Only a server-side diff covers both paths.

**The render-time write is deliberate.** It closes the gap between the previous page's `pagehide` DELETE and the next tick (`includes/heartbeat.php:88-90`). Removing it makes presence stale during in-admin navigation.

#132 should measure both paths separately: renders scale with navigation, ticks with concurrent sessions.

josephfusco:
This issue was identified as part of a peer review conducted by Claude Sonnet 4.6, facilitated by @josephfusco.

jigneshbhavani:
I'd like to take this one.

The write table needs updating since #284 landed. The `lock-{id}` row is gone, so the post-edit tick is 2 writes and not 3, and the lines have moved: render at `includes/heartbeat.php:117`, post-edit render `:124`, tick `:220`, post-edit tick `:256`. All four go through `wp_set_presence()` (`includes/functions.php:95`), so the server-side diff has one place to live.

I'd do it inside the existing statement rather than reading before writing, making `date_gmt` conditional on the data having changed or the row being past a refresh threshold. Measured on MariaDB, an unchanged and still-fresh tick then reports 0 affected rows.

Worth your call: on SQLite the same statement stores the right value but still writes the row every time, so Playground keeps correct behaviour without the saving. Fine by me given the target is cheap MySQL hosting, but say if you'd rather the engines not diverge.

[...]

PR #284, opened 2026-08-14
fix: merge the post lock entry into the editor's presence entry

### LINKED ISSUE #134: Resolve double-entry behavior in post editor rooms
When a user opens a post editor, two presence entries are written to the same post room:

- `editor-{user_id}` by the editor heartbeat handler (`includes/heartbeat.php:108-116`)
- `lock-{user_id}` by the post-lock bridge (`includes/post-lock-bridge.php:49-56`)

Both store identical data: same room, same user, same `{ action: 'editing', screen: ... }`. They differ only in the `client_id` prefix. `wp_get_presence()` returns both, undocumented, with no way to tell them apart except by parsing that prefix.

The one signal the split encodes is post-lock liveness, since the two rows are refreshed by different heartbeat keys and age out independently.

## Proposed fix

Merge into one `editor-{user_id}` entry with a `locked` boolean in the state data. The bridge writes into the existing entry instead of creating a row.

Knock-ons:

- Resolves #21. With no separate lock row there is nothing to keep alive in a hidden tab, and PR #78 closes.
- Removes one write per tick per editing user, changing the baseline #132 should benchmark.
- `includes/heartbeat.php:80-83` registers `lock-{user_id}` for the `pagehide` DELETE. That goes away.

## If rejected

#21 is a live bug, PR #78 is its fix and should be unblocked. This issue then narrows to adding an explicit `type` field so callers stop parsing `client_id`.

## Sequencing

#135 moves this same inline script to a file. Land it after this and PR #78.

jigneshbhavani:
Opening a post editor wrote two rows to the same post room for the same user, `editor-{user_id}` from the editor heartbeat and `lock-{user_id}` from the post-lock bridge. Both stored identical state and differed only by a `client_id` prefix that nothing in the plugin parses.

The lock is now a `locked` flag on the editor's own entry, read from the `wp-refresh-post-lock` payload that arrives on the same tick. That keeps the one signal the split encoded: a tick carrying no refresh clears the flag, which is what the second row ageing out used to say.

Worth being explicit, since the flag is easy to over-read. `locked` means this user's editor is sending lock refreshes, not that they hold the lock. Core sends `wp-refresh-post-lock` from any open editor, including one showing the takeover dialog. That is what the `lock-` row meant too, so this preserves the meaning rather than changing it.

## The bridge stands down instead of writing

`wp_presence_bridge_post_lock()` returns early when `presence-editor-ping` covers the same post, since the editor handler has already written that row from the same payload. That removes a write per tick per editing user rather than relocating one, which is the baseline change #131 is waiting on. The bridge still writes when a client refreshes the core lock without sending the presence ping. `heartbeat.php` also drops `lock-{user_id}` from the pagehide cleanup, since that client no longer exists.

## Second commit: the widget counts people, not rows

Nothing guarantees one row per user in a room, and the Active Posts widget was the only aggregation path that did not already dedupe. `wp_get_active_rooms()`, `post-list.php:90` and `admin-bar.php:109` all key by user id. It also covers the upgrade window: `lock-*` rows written by the old code survive up to the TTL after deploy and the pagehide cleanup no longer targets them, so the dedupe is what keeps the dashboard honest until they expire.

## Testing

`wp_presence_editor_heartbeat_received()` had no coverage, so this adds it alongside the new behaviour, plus the regression guard that one editing user occupies one row however many handlers see the tick. Every guard was mutation checked, including restoring the old `lock-` write, which fails that guard.

Verified in a browser on wp-env against both branches: on `main` the room holds `editor-1` and `lock-1` and the dashboard reads "2 editors" for one person; on this branch the same steps give one row with `"locked":true` and the widget agrees with `GET /presence/rooms` reporting `user_count: 1`. Note the post lock heartbeat only runs when a site has more than one user, because core hooks the lock dialog under `get_user_count() > 1` (`edit-form-advanced.php:28`), so on a single user site neither row is written and the bug is invisible.

205 tests, 5 skipped (multisite only). PHPCS, PHPStan and Playwright clean.

Fixes #134

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5
Used for: Tracing the consumers of presence rows, implementing the merge, and writing the tests. I confirmed each test fails before the change and passes after, mutation checked every regression guard, verified the behaviour in a browser against both branches, and ran the full PHPUnit and Playwright suites plus PHPCS and PHPStan.

### DIFF
--- a/includes/heartbeat.php
+++ b/includes/heartbeat.php
@@ -10,6 +10,25 @@ if ( ! defined( 'ABSPATH' ) ) {
 }
 
 /**
+ * Builds the state stored on a user's entry in a post room.
+ *
+ * Shared so the two writers of that entry cannot drift apart on its shape.
+ *
+ * @access private
+ *
+ * @param string $screen_id The screen ID.
+ * @param bool   $locked    Whether this write carries a post lock refresh.
+ * @return array The state to store.
+ */
+function wp_presence_editor_state( $screen_id, $locked ) {
+	return array(
+		'action' => 'editing',
+		'screen' => $screen_id,
+		'locked' => (bool) $locked,
+	);
+}
+
+/**
  * Enqueues heartbeat and the presence ping script on all admin pages.
  */
 function wp_presence_enqueue_heartbeat_ping() {
@@ -75,11 +94,6 @@ function wp_presence_enqueue_heartbeat_ping() {
 						'room'      => $room,
 						'client_id' => 'editor-' . $user_id,
 					);
-					// The post-lock bridge writes this entry via the wp-refresh-post-lock heartbeat.
-					$entries[] = array(
-						'room'      => $room,
-						'client_id' => 'lock-' . $user_id,
-					);
 				}
 			}
 		}
@@ -105,13 +119,12 @@ function wp_presence_enqueue_heartbeat_ping() {
 	if ( $editor_post_id ) {
 		$editor_room = wp_presence_post_room( $editor_post_id );
 		if ( $editor_room ) {
+			// No tick has carried a lock refresh yet. connectNow() on load makes
+			// that a single request, not a visible state.
 			wp_set_presence(
 				$editor_room,
 				'editor-' . $user_id,
-				array(
-					'action' => 'editing',
-					'screen' => $screen_id,
-				),
+				wp_presence_editor_state( $screen_id, false ),
 				$user_id
 			);
 		}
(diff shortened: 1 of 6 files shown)


josephfusco (approved):
Ran this locally in a worktree against both branches — PHPCS/PHPStan clean, full suite passes (205/205, 5 skipped), and the one multisite failure I saw also shows up on unmodified main so it's not a regression; walked through the bridge's stand-down logic and the widget's dedupe by hand and both check out.

Thank you for working on this!

PR #288, opened 2026-08-15
perf: skip Who's Online payload when room state is unchanged

### LINKED ISSUE #268: Skip the response body when presence state hasn't changed
Return a cheap "nothing changed" signal when a room's state hash matches what the client already has, instead of rebuilding and shipping the full JSON payload every tick. Standard conditional-request pattern (ETag-equivalent), applied to the Heartbeat response shape rather than a full HTTP 304, given Heartbeat's single-request-many-features design.

Part of #267.

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Sonnet 5

i-am-chitti:
### Description

Closes #268.

Server hashes the room's meaningful state; the client echoes its last hash. On a match the response carries last-seen timestamps instead of the payload, skipping `cache_users()`, per-user `get_userdata()` / `get_avatar_url()`, label building, and most of the JSON.

### Review notes

- The unchanged reply carries a `user_id → last-seen` map, not a bare flag: the hash excludes `date_gmt` (it would flip every tick), so a flag alone breaks the idle dot for backgrounded tabs. Verified the dot still greys at 30s and clears at 60s, matching `main`.
- Display names and avatars aren't on the presence row, so profile changes won't reach other widgets until room state changes for another reason.

### Testing

1. Log in as two users - admin in a normal window, an editor in an incognito window - both on the Dashboard. Position the windows side by side so neither covers the other; a fully covered window reports itself hidden and stops pinging, which invalidates the test.
2. Confirm the editor appears in the admin's Who's Online widget with a solid green dot.
3. In the admin window, open DevTools → Network and filter `admin-ajax` (presence rides WordPress's Heartbeat; it has no endpoint of its own). After the first tick, responses should carry `presence-online-unchanged` with a `user_id → last-seen` map instead of the full `presence-online` array, and drop noticeably in size.
4. In the editor window, switch to a second tab so the Dashboard is backgrounded. Leave it open - closing it deletes the presence entry and tests nothing.
5. Watch the editor's row in the admin window. The dot should turn grey at ~30s and the row should clear at ~60s - identical to `main` - while the admin's responses keep coming back as `presence-online-unchanged`.

### Out of scope

`heartbeat_received()` runs on every admin screen but only the Dashboard consumes the data - off-Dashboard ticks still build an unread payload and can't be hash-suppressed. Will be raising a issue for this.

### Measurement

To follow

### Screenshots

image

image

image

### Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Sonnet 5
Used for: Implementation and tests

### DIFF
--- a/includes/widgets/class-wp-presence-widget-whos-online.php
+++ b/includes/widgets/class-wp-presence-widget-whos-online.php
@@ -733,7 +733,58 @@ JS,
 		}
 
 		$entries = wp_get_presence( wp_presence_admin_room() );
-		$online  = array();
+		$hash    = self::hash_online_state( $entries );
+
+		$client_hash = isset( $data['presence-online-hash'] ) ? sanitize_text_field( $data['presence-online-hash'] ) : '';
+
+		if ( $client_hash && $client_hash === $hash ) {
+			$response['presence-online-unchanged'] = true;
+
+			return $response;
+		}
+
+		$response['presence-online']      = self::build_online_entries( $entries );
+		$response['presence-online-hash'] = $hash;
+
+		return $response;
+	}
+
+	/**
+	 * Hashes the meaningful state of a room's presence entries.
+	 *
+	 * Excludes date_gmt, which every tick rewrites for the pinging user and so
+	 * would flip the hash on every tick.
+	 *
+	 * @param array $entries Presence entry objects from wp_get_presence().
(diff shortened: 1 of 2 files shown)


josephfusco (approved):
Looks good to me!

Issue #122, opened 2026-07-25
Add usePresenceUsers React hook for polling presence rooms


josephfusco:
Originated from a [#feature-presence-api](https://wordpress.slack.com/archives/C0ASCNMB7FZ/p1776358400206169) Slack thread where @sethrubenstein and @benwormald built a `usePresenceFelt` hook to detect if a user is editing an entity directly and lock editing when using `useEntityBlockEditor`. @benwormald is using it to detect when a designer is working on a chart block and prevent other users from editing it while they are active, surfacing a "Benjamin Wormald is editing this chart. Changes are disabled." banner in the editor.

<img width="1293" height="727" alt="Screenshot of block editor showing 'Benjamin Wormald is editing this chart. Changes are disabled.' " src="https://github.com/user-attachments/assets/167aa70a-7e25-413c-a1fc-c3d62299bd86" />

```js
/**
 * WordPress Dependencies
 */
import { useState, useEffect, useRef, useCallback } from '@wordpress/element';
import { useSelect } from '@wordpress/data';
import apiFetch from '@wordpress/api-fetch';

/**
 * Default interval (ms) for Presence API REST fetches.
 * Matches the default WordPress Heartbeat interval (15 s).
 */
const DEFAULT_INTERVAL_MS = 15000;

/**
 * Polls the WordPress Presence API for occupants of a room and resolves display names.
 *
 * When `room` is null or undefined, no requests run and the hook reports empty presence
 * (nothing to observe).
 *
 * @param {string|null|undefined} room                        Presence room id (e.g. `postType/chart:123`).
 * @param {Object}                [options]
 * @param {boolean}               [options.includeSelf=false] When false, the current user is
 *                                                            excluded from `users`.
 * @param {number}                [options.interval=15000]    Fetch interval in ms.
 *
 * @return {{
 *   isPresent: boolean,
 *   users: Array<{ userId: number, displayName: string }>
 * }}
 */
export default function usePresenceFelt(room, options = {}) {
	const { includeSelf = false, interval = DEFAULT_INTERVAL_MS } = options;

	const [userIds, setUserIds] = useState([]);

	const currentUserId = useSelect(
		(select) => select('core').getCurrentUser()?.id,
		[]
	);

	const userNameCache = useRef({});
	const prevUserIdKey = useRef('');
	const intervalRef = useRef(null);

	const fetchPresence = useCallback(async () => {
		if (!room) {
			return;
		}

		try {
			const entries = await apiFetch({
[...]

PR #292, opened 2026-08-16
feat: add usePresenceUsers React hook

josephfusco:
Implements the React hook from #122. Ships as source (no build step) following WordPress/Gutenberg pattern.

Addressed Gutenberg peer review feedback: jQuery elimination, TypeScript, tests, accessibility, ref-based effect optimization.

**Used for**: Assisting with implementation and peer review


josephfusco:
Addressed all issues from Gutenberg peer review:

**Critical**: jQuery dependency (abstraction layer), race condition (fetchInProgress ref), unstable deps (ref pattern, only re-subscribes on room change).

**Major**: TypeScript defs, peerDependencies, 11 unit tests (all passing).

**Minor**: Loading state, error handling, accessibility docs, naming (`userId` → `id`).

**Build**: Removed. Ships as source following WordPress/Gutenberg pattern - consumers import and bundle themselves.

**Used for**: Assisting with review implementation

Issue #298, opened 2026-08-18
feat: network presence


josephfusco:
Adds a network-wide presence summary. `wp_presence` is per-site; nothing aggregates it across a network today.

Each site pushes its own online count into a small network-level table when its own presence data changes, instead of a central process reading every site's table on demand. Reading the summary is one query against that table, not a fan-out across every site's `wp_presence` table.

This avoids `switch_to_blog()` entirely, since no per-site table is ever queried at read time. It also avoids the write contention a single shared option or transient would create as the network grows: different sites write different rows, so writes from different sites never block each other.

Shows up in the Sites list, Users list, and a network dashboard widget matching the existing single-site Who's Online widget.

The read path needs no large-network cutoff, since it is always a single query against a table sized to the network rather than one query per site. The write path does, and does not have one yet: see #348.

## Status

Six pull requests, merged bottom-up. Each one retargets to `main` as the one below it lands.

- [ ] #332 summary table
- [ ] #333 write path
- [ ] #334 read path
- [ ] #335 Online column on Network Admin > Sites
- [ ] #336 Online view and column on Network Admin > Users
- [ ] #337 Who's Online widget on the Network Admin dashboard

Nothing is user-visible until #335. The three below it add a table, write to it, and expose functions that read it, with no admin surface.

## Before this ships

- [ ] #348 large-network gate on the push, and a way to turn aggregation off
- [ ] #349 end-to-end coverage for the three Network Admin screens

## After this ships

Small, self-contained, and each one claimable on its own. Tracked under the `Multisite polish` milestone: #328, #329, #330, #331, #343, #344, #351.

Open design questions, not blocking: #311 neutral avatar state, #312 avatar-stack CSS and JS duplication.

<details>
<summary>Use of AI Tools</summary>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

Used for: Assisting with the status breakdown

</details>

josephfusco:
Shipped in 0.2.0. #349 moved to 0.2.1. #329, #344 and #351 remain open on their own milestones.

PR #299, opened 2026-08-18
feat: network-wide presence aggregation for multisite

### LINKED ISSUE #310: Network presence push writes to one shared global table
#299 replaces the per-request UNION across each site's presence table with a push model: each site upserts its online-user row into one shared global table (wp_presence_network_summary), keyed by blog_id.

The write is coalesced, so a tick that finds the same people online is a zero-row no-op and the row is only rewritten when the user set changes or the row goes stale (TTL minus the idle heartbeat interval, 15s at the defaults). That bounds the rate but not the concentration.

Per-site tables can be split across database shards on a large network. A global table can't be split the same way, so every site's write still lands on one table instead of its own shard.

Does that write concentration hold up at large-network scale, or does the write path need to stay shardable?

PR: #299

### LINKED ISSUE #318: Two network helpers land in #299 with no callers until #313
`wp_presence_render_avatar_stack()` (`includes/functions.php:878`) and `wp_presence_get_network_online_user_ids()` (`includes/network-functions.php:250`) are added in #299 but every caller lives in #313, so merging #299 on its own puts two unused functions on main.

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

Used for: Assisting with an unused-function audit

</details>

### LINKED ISSUE #321: Network summary is rebuilt once per list table row
`wp_presence_get_network_summary()` has no memoization and no object cache, so the Sites and Users list columns rebuild the whole summary once per row.

Both list-table docblocks claim the opposite:

> wp_presence_get_network_summary() is one query against the shared network summary table, not one query per site, so this costs one query per page load rather than one per row.

Measured on multisite with 8 seeded sites, calling `wp_presence_render_network_sites_column()` once per row:

```
```

The query is the cheap part. Each call also decodes every row's JSON and runs `get_userdata()` and `get_avatar_url()` per user. At 300 sites averaging 15 users that is 135ms per build, so roughly 2.7s on a 20 row Sites screen. Network Users is worse, since `wp_presence_get_network_online_user_ids()` rebuilds it for the view count and the query filter rebuilds it again.

Fix: memoize in a static and cache the built summary in the object cache, and replace the per-row `get_site()` with a single `_prime_site_caches()`.

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

Used for: Assisting with review and measurement

</details>

### LINKED ISSUE #324: Network summary read path is unbounded
`wp_presence_compute_network_summary()` selects every row above the cutoff with no `LIMIT`, then hydrates `display_name` and `get_avatar_url()` for every user on every site. Cost is linear in total users online network-wide, and it is recomputed per heartbeat tick per viewing admin.

Measured on wp-env multisite with no persistent object cache, `wp_presence_compute_network_summary( 60 )` from a cold static:

| network | online | time | queries | serialized size |
|---|---|---|---|---|
| 50 sites x 5 | 250 | 12 ms | 2 | 49 KB |
| 300 sites x 10 | 3,000 | 146 ms | 2 | 557 KB |
| 1000 sites x 15 | 15,000 | 599 ms | 2 | 2.7 MB |

Two queries at every size, so the table design holds. The cost is all PHP hydration above it.

No consumer needs the full set. The Sites list column renders an avatar stack of the first few per site, the Users list column needs site names only, and the widget truncates. Fix is to bound the query and the hydration to what the caller asked for, and to count in SQL rather than by hydrating and calling `count()`.

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

</details>

### LINKED ISSUE #326: Network summary rows are never removed on site delete or uninstall
Nothing deletes a site's row from `{base_prefix}presence_network_summary`. `wp_delete_site()` has no handler, and `uninstall.php` drops the per-site `presence` tables and options but never the summary table, `wp_presence_network_summary_db_version`, or the `wp_presence_network_summary_table.lock` site option.

`wp_presence_compute_network_summary()` skips rows whose `blog_id` no longer resolves, so the effect is dead rows accumulating in a table pinned to the global cluster rather than wrong output.

Fix is a `wp_delete_site` handler for the row and summary-table cleanup in `uninstall.php`.

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

</details>

josephfusco:
Multisite networks had no way to see who's online across sites.

Part of #298
Closes #310
Closes #318
Closes #321
Closes #324
Closes #326

<details>

AI assistance: Yes
Tool(s): Claude Code

</details>


josephfusco:
Went through this against the sharding question in #310, with the numbers measured on a live multisite rather than read off the diff.

### Write amplification on every pageview

`wp_set_presence()` goes from 1 query to 4 with the push hook active:

```
wp_set_presence queries with push=4 without push=1
```

This is not just the heartbeat tick. `wp_presence_enqueue_heartbeat_ping()` writes the admin room on every admin pageview and every front-end view with an admin bar, so every pageview by every `edit_posts` user on every site now issues a SELECT plus an INSERT ... ON DUPLICATE KEY UPDATE against `{base_prefix}presence_network_summary`.

Coalescing bounds row churn. It does not bound statement rate, connection count, or row-lock acquisition, and the SELECT that feeds the comparison runs either way. Full answer to #310 over on that issue.

Fix: debounce the push through the object cache so a site pushes at most once per `wp_presence_network_summary_refresh_interval()`, and default the aggregation off under `wp_is_large_network()`.

### Nothing ever removes a row

There is no `wp_uninitialize_site` or `wp_delete_site` hook, so a deleted site's row stays forever. The read path skips it because `get_site()` returns null, so this is a leak rather than a correctness problem.

`uninstall.php` also never drops `{base_prefix}presence_network_summary` and never deletes `wp_presence_network_summary_db_version` or `wp_presence_network_summary_table.lock`.

### The read path is unbounded

`wp_presence_compute_network_summary()` selects every row over the cutoff with no LIMIT, hydrates every user on every site, then sorts twice. `get_site()` runs per row with no `_prime_site_caches()`.

At 300 sites averaging 15 users, the avatar hydration alone:

```
300 sites x 15 users get_avatar_url: 135ms per summary build
```

### Smaller

- `JSON_PRETTY_PRINT` inflates the row 2.8x, 241 bytes against 86 for 15 users. That column is the operand of `data <> VALUES(data)` on a longtext, so the inflation is paid on every push comparison and every read.
- `false !== $result` treats a delete matching zero rows as a change, so the pagehide path fires spurious pushes.
- `url` is built from `is_ssl()` on the current request, so a mixed-scheme network gets wrong links for some sites.
- `wp_presence_register_network_summary_table()` appends to `ms_global_tables` with no `in_array` guard and runs both at require time and on `init` priority 0, so the entry lands twice.
- `wp_presence_network_summary_table_exists()` runs `SHOW TABLES LIKE` on every `admin_init` once provisioned. Same shape as the single-site table, but a network with tens of thousands of tables turns it into a data dictionary scan per admin request.

josephfusco:
Closing in favour of the stack this was split into: #332, #333, #334, #335, #336, #337.

Same code. Every production file's token stream is identical to this branch; the only difference is that two functions moved to the layer they belong to. What changed is the tests, which now share one `WP_Presence_Network_UnitTestCase` instead of repeating the same scaffolding three times.

Each pull request answers one reviewer question rather than six. The review notes from this thread are re-posted on the one they belong to.

PR #300, opened 2026-08-19
perf: coalesce presence polling across tabs of the same user

### LINKED ISSUE #269: Coalesce polling across tabs of the same user
When the same user has the same admin screen open in multiple tabs, each tab independently polls. Use `BroadcastChannel` to elect one tab as the poller and share results with the others client-side — no server changes, pure client-side request reduction.

Part of #267.

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Sonnet 5

i-am-chitti:
Fixes #269

## Summary

When the same user has an admin screen open in multiple tabs, each tab independently pings Heartbeat, triggering its own `wp_set_presence()` write and Who's Online/Active Posts/stale-screen lookup - redundant work that scales with tab count, not with actual viewers.

Elects one tab per context (Web Locks) to send the ping; sibling tabs skip their own ping and instead receive the elected tab's response over `BroadcastChannel`, so their widgets/banners stay live. Closing the elected tab promotes a sibling automatically. Tabs on a different post or screen are never coalesced with each other.

Covers all three tab-scoped ping mechanisms in the plugin:

- `presence-ping` / `presence-editor-ping` (admin + editor rooms, Who's Online) - dedupe key is screen + post + front-page context
- `presence-screen-ping` (stale-screen banner) - dedupe key is `screenKey`
- `presence-active-posts-ping` (Active Posts widget) - fixed dedupe key, since this ping carries no per-page context (sitewide data). Required first extracting its script from inline PHP into an external file (`assets/js/active-posts.js`) so it could be enqueued and coalesced like the other two.

Also ships `usePresenceUsers`, a React hook + coordinator for third-party consumers of the public presence REST endpoint - a related but separate improvement, unaffected by the wp-admin fix above since nothing in this plugin's own UI uses that hook yet.

## Test plan

- [x] New unit tests for the coordinator (`presence-poll-coordinator.test.js`)
- [x] New e2e tests per mechanism: election, promotion on leader-tab close, response relay, different-post/screen isolation
- [x] `npm run test` - no regressions
- [x] `npm run test:e2e` - no regressions
- [x] Manual multi-tab testing across all three mechanisms

<details>
<summary>Screenshots</summary>

image

image

image

image

image

</details>

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Sonnet 5
Used for: Assisting with implementation

</details>

### DIFF
--- a/assets/js/active-posts.js
+++ b/assets/js/active-posts.js
@@ -16,32 +16,7 @@
 	// Sitewide data, no per-tab context — the dedupe key is fixed.
 	var pingContextKey = 'wp-presence-active-posts-ping';
 
-	var hasLocks = typeof navigator !== 'undefined' &&
-		navigator.locks &&
-		typeof navigator.locks.request === 'function';
-
-	// No Locks API: ping independently, same as before.
-	var isPingLeader = ! hasLocks;
-
-	var pingChannel = typeof BroadcastChannel === 'function'
-		? new BroadcastChannel( pingContextKey )
-		: null;
-
-	if ( pingChannel ) {
-		pingChannel.addEventListener( 'message', function ( event ) {
-			$( document ).trigger( 'heartbeat-tick', [ event.data ] );
-		} );
-	}
-
-	if ( hasLocks ) {
-		// All tabs queue on this lock; the winner leads until its tab closes.
-		navigator.locks
-			.request( pingContextKey, function () {
-				isPingLeader = true;
-				return new Promise( function () {} );
-			} )
-			.catch( function () {} );
-	}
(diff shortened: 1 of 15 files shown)


josephfusco:
The only concern from review was that the tab-election/relay logic was duplicated across presence-ping.js, active-posts.js, and stale-screen.js, so I pushed a commit extracting it into a shared tab-coordinator.js helper, verified with the existing coalescing e2e tests plus the full PHPUnit/PHPCS/PHPStan suite.

josephfusco (approved):
Thanks for knocking this out, nice work with this!

PR #313, opened 2026-08-22
feat: network presence UI for the Sites list, Users list, and network dashboard

### LINKED ISSUE #298: feat: network presence

### LINKED ISSUE #322: Network Users Online view is overwritten by the single-site pre_get_users filter
The Network Users "Online" view lists users online on the current site, not users online anywhere on the network.

`wp_presence_filter_network_online_users()` sets `include` on `users_list_table_query_args`. `WP_MS_Users_List_Table::prepare_items()` then instantiates `WP_User_Query` (`wp-admin/includes/class-wp-ms-users-list-table.php:96`), which fires `pre_get_users`, where the single-site `wp_presence_filter_online_users()` is hooked. Its only guards are `is_admin()` and `edit_posts`, both true for a super admin in network admin, so it overwrites `include` with the current site's online users and appends the viewer's own ID.

Fix: guard `wp_presence_filter_online_users()` with `! is_network_admin()`.

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

</details>

### LINKED ISSUE #325: Network widget heartbeat sends the full summary whenever anything on the network changes
`WP_Presence_Network_Widget_Whos_Online::heartbeat_received()` hash-gates the payload, but the gate does not hold at scale and the build cost is paid either way.

`hash_summary()` covers blog_id and user IDs across the entire network, so a single arrival or departure among any of the online users invalidates it. On a large network the gate essentially never matches, and every tick falls through to the full send.

The send is unsliced. `render()` displays `self::VISIBLE_SITES` sites with `wp_presence_render_avatar_stack()` capping each at 4 avatars, while the tick assigns `$response['presence-network-widget'] = $summary['sites']`, which at 1000 sites x 15 users is 2.7 MB.

`wp_presence_get_network_summary()` is also called before the hash comparison, so the 599 ms build in #324 happens on every gated tick whether the hash matches or not.

Fix is to send what the widget renders, sliced to `VISIBLE_SITES` and capped at the avatar-stack maximum, and to hash that same slice so a change on an unrendered site stops invalidating it.

Depends on #324 for the bounded read.

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

</details>

josephfusco:
The network summary from #299 had no surfaces reading it.

Depends on #299 and targets its branch; will retarget to main once #299 merges.

Fixes #298
Closes #322
Closes #325

<details>

AI assistance: Yes
Tool(s): Claude Code

</details>

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -822,3 +822,29 @@ function wp_presence_hydrate_room_users( $rooms, $timeout = WP_PRESENCE_DEFAULT_
 
 	return $rooms;
 }
+/**
+ * Renders a small avatar stack for a list of users.
+ *
+ * Shared across every surface that shows an overlapping avatar stack (the
+ * dashboard widget's overflow indicator, the network Sites list column, the
+ * network dashboard widget) so they all render the stack identically.
+ *
+ * @access private
+ * @param array $users Users, each with 'avatar_url' and 'display_name'.
+ * @param int   $max   Optional. Maximum avatars to show. Default 4.
+ * @param int   $size  Optional. Avatar width and height in pixels. Default 20.
+ * @return string HTML markup.
+ */
+function wp_presence_render_avatar_stack( $users, $max = 4, $size = 20 ) {
+	$stack_max = min( count( $users ), $max );
+	$html      = '<span class="presence-avatar-stack">';
+
+	foreach ( array_slice( $users, 0, $stack_max ) as $index => $user ) {
+		$z     = $stack_max - $index;
+		$html .= '<img src="' . esc_url( $user['avatar_url'] ) . '" width="' . (int) $size . '" height="' . (int) $size . '" style="z-index:' . (int) $z . '" alt="' . esc_attr( $user['display_name'] ) . '" />';
+	}
+
+	$html .= '</span>';
+
+	return $html;
+}
--- a/includes/network-functions.php
+++ b/includes/network-functions.php
@@ -16,6 +16,13 @@ if ( ! defined( 'ABSPATH' ) ) {
 	exit;
 }
 
+// How many people a network surface draws per site. Every one of them costs a
+// user load and an avatar URL, so this is also what the read path is asked to
+// resolve; the headcount beside the stack is the site's real total.
+if ( ! defined( 'WP_PRESENCE_NETWORK_AVATARS' ) ) {
+	define( 'WP_PRESENCE_NETWORK_AVATARS', 4 );
+}
+
 /**
  * Checks whether the network-wide presence summary table exists.
  *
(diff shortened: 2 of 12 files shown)


josephfusco:
Two blockers, both measured on a live multisite, filed as #321 and #322.

### The summary is rebuilt once per row

#321. `wp_presence_get_network_summary()` has no memoization and no object cache, so the Sites and Users columns rebuild it per row. Both list-table docblocks claim the opposite.

```
```

The query is the cheap part. Each call also decodes every row's JSON and runs `get_userdata()` and `get_avatar_url()` per user, 135ms per build at 300 sites averaging 15 users.

### The network Online filter returns the wrong users

#322. `wp_presence_filter_network_online_users()` sets `include` on `users_list_table_query_args`, then `WP_MS_Users_List_Table::prepare_items()` builds a `WP_User_Query` that fires `pre_get_users`, where the single-site filter overwrites it with the current site's users.

### Heartbeat ships the whole summary

`heartbeat_received()` returns `$summary['sites']` with avatar URLs for every user on every site, every tick, to every network admin with the dashboard open. The widget renders `VISIBLE_SITES = 5`.

### Smaller

- `hash_summary()` covers blog IDs and user IDs only, so a display name or avatar change never refreshes the widget.
- `buildAvatarStack()` in the inline JS skips users with no `avatar_url` and `wp_presence_render_avatar_stack()` does not, so the server and client renders differ.

josephfusco:
Closing in favour of the stack this was split into: #332, #333, #334, #335, #336, #337.

Same code. Every production file's token stream is identical to this branch; the only difference is that two functions moved to the layer they belong to. What changed is the tests, which now share one `WP_Presence_Network_UnitTestCase` instead of repeating the same scaffolding three times.

Each pull request answers one reviewer question rather than six. The review notes from this thread are re-posted on the one they belong to.

PR #323, opened 2026-08-24
perf: push the network summary only when the site's online set changes

### LINKED ISSUE #310: Network presence push writes to one shared global table

josephfusco:
The summary table is pinned to the global cluster on a sharded network, and #299 pushed to it on every admin pageview.

Depends on #299 and targets its branch.

Closes #310

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

</details>

### DIFF
--- a/includes/network-functions.php
+++ b/includes/network-functions.php
@@ -138,18 +138,17 @@ function wp_presence_network_summary_refresh_interval() {
  * Pushes the current site's online-user snapshot into the network-wide
  * summary table.
  *
- * Runs on wp_presence_admin_room_changed, so every admin-room write pushes,
- * not just the heartbeat tick: logout and the pagehide delete clear the site
- * immediately instead of waiting for its entry to age out.
+ * Runs on wp_presence_admin_room_changed, so an arrival or a departure reaches
+ * the network within the request that caused it: logout and the pagehide delete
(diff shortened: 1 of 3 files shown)


josephfusco:
Measured on live multisite, queries per admin-room write:

| | before | after |
|---|---|---|
| arrival | 4 | 6 |
| steady-state tick | 4 | 2 |
| departure | 4 | 5 |

Arrival and departure each pay one extra write to record what was pushed, and the tick that used to reach the global cluster now stops at an autoloaded option. The one query left on a tick is the `wp_get_presence()` read that builds the payload, which is on the site's own shard.

Issue #330, opened 2026-08-24
ms_global_tables gets a duplicate presence_network_summary entry


josephfusco:
`wp_presence_register_network_summary_table()` appends to `$wpdb->ms_global_tables` with no `in_array()` guard and runs twice, at require time in `presence-api.php` and again on `init` priority 0, so the array holds `presence_network_summary` twice on every multisite request.

```php
if ( ! in_array( 'presence_network_summary', $wpdb->ms_global_tables, true ) ) {
	$wpdb->ms_global_tables[] = 'presence_network_summary';
}
```

`wpdb::tables()` rekeys by table name when `$prefix` is true, so the duplicate only survives in the raw array and in `tables( $scope, false )`, which backup and migration plugins read directly. The per-site half of the same bug was fixed in #345.

jigneshbhavani:
I'd like to take this one.

The same bug is already on main. `wp_presence_register_table()` appends to `$wpdb->tables` with no guard and is called the same two ways, at require time (`presence-api.php:71`) and on `init` priority 0 (`:281`), so `$wpdb->tables` holds `presence` twice on every install, single site included. Measured on a running site: 2 before, 1 after the guard. The network version copied a pattern that was only safe because the per-site one assigns a property, and the append is what breaks it.

Two things in the issue body don't hold up, and they narrow the fix rather than widen it.

`set_blog_id()` never sees the duplicate. It iterates `tables( 'blog' )` and `tables( 'old' )`, and `'blog'` is `$this->tables` alone while `'old'` pulls `old_ms_global_tables`, so `ms_global_tables` is in neither.

`tables()` hides it on the default path. With `$prefix = true` the loop rekeys each entry as `$tables[ $table ] = $prefix . $table`, so duplicates collapse. Measured on multisite with the summary table registered:

| scope | `tables($scope)` | `tables($scope, false)` |
|---|---|---|
| all | 1 | 2 |
| global | 1 | 2 |
| ms_global | 1 | 2 |
| blog | 0 | 0 |

The only core caller passing `false` is `update.php:131`, and it breaks on the first match, so nothing in core behaves differently either way. What is left is the raw arrays, which backup and migration plugins read directly, and the contract itself.

I'd guard both appends rather than drop one of the two call sites. Both call sites date to 9c55888 with no recorded reason for the pair, so removing one is a load-order guess, and making the function idempotent fixes the class instead of one instance.

[...]

PR #332, opened 2026-08-24
feat: add a network-wide presence summary table

### LINKED ISSUE #326: Network summary rows are never removed on site delete or uninstall

josephfusco:

1. **#332 summary table (this one)**
2. #333 write path
3. #334 read path
4. #335 Sites list column
5. #336 Users list

---

Reading who is online across a network means reading every site's presence table, and there is no way to do that from one site without switching into each of them.

Part of #298
Closes #326

<details>

AI assistance: Yes
Tool(s): Claude Code

</details>

### DIFF
--- a/includes/network-functions.php
+++ b/includes/network-functions.php
@@ -1,0 +1,136 @@
+<?php
+/**
+ * Presence API network functions.
+ *
+ * Reading who is online across a network means reading every site's presence
+ * table, and there is no way to do that from one site without switching into
+ * each of them. This file provisions the alternative: one shared, network-wide
+ * summary table holding a row per site, so the network can be read with a
+ * single query against a single table.
+ *
+ * The table is registered as an ms_global_tables entry on $wpdb->base_prefix
+ * (see wp_presence_register_network_summary_table() in presence-api.php), the
+ * same way core registers blogs and sitemeta, so it is created once per
+ * network rather than once per site.
+ *
+ * @package Presence_API
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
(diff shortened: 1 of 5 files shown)


josephfusco:
Review findings from #299 that land here are #326, closed by this one, and #330, which is not.

josephfusco:
Merge order note, since the stack cannot go in as-is.

CI and CodeQL filter `pull_request` to `branches: [ main ]`, so #333 through #337 run 4 checks each against the 17 here. #340 removes the filter.

For a `pull_request` event GitHub reads the workflow file from the merge commit, so merging #340 into `main` is not enough on its own; each stack branch has to carry it. Sequence:

1. Merge #340.
2. Rebase this branch onto `main`, then each branch above it onto its parent, bottom-up, pushing with `--force-with-lease`.
3. Full CI then runs on all six, and they merge bottom-up from here.

Nothing else in the stack is waiting on anything.

Issue #329, opened 2026-08-24
Network summary rows are stored pretty-printed, at 2x the bytes


josephfusco:
`wp_presence_encode_network_summary_row()` writes the `data` column with `JSON_PRETTY_PRINT`, so every user ID gets its own indented line. The read path selects that column in full for every site on the network on each network-admin request.

Measured on a seeded network of 1000 sites x 15 users:

| | `SUM(LENGTH(data))` | cold read |
|---|---|---|
| `JSON_PRETTY_PRINT` (current) | 306 KB | 38.4 ms |
| compact | 156 KB | 39.4 ms |

Twice the stored and transferred bytes, with no latency difference over a local socket. The read cost is elsewhere, so this is a storage and DB-link question rather than a latency one, and it only starts to matter on a network large enough to have a remote database.

The flag is deliberate: `test_row_data_is_readable_on_its_own` asserts the row is legible when someone reads the table directly. Worth deciding whether that debuggability is worth 2x on a column read in full on every request, rather than leaving it as an unexamined default.

<details>
<summary>Use of AI Tools</summary>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5
Used for: Assisting with analysis

</details>

PR #333, opened 2026-08-24
feat: push each site's online set into the network summary

### LINKED ISSUE #327: Deleting a user leaves their presence behind on every site they were online on
Nothing in the plugin hooks user deletion. `grep -rn "deleted_user\|wpmu_delete_user\|remove_user_from_blog" includes/ presence-api.php` returns nothing; `includes/lifecycle.php:40` is the logout handler, not a delete handler.

A deleted user's rows in each site's `{prefix}presence` table survive until they age out on TTL, and that site's network summary row keeps naming the ID until its next push. So for up to one TTL the network reports a headcount that includes an account that no longer exists.

The read path compensates. `wp_presence_filter_network_snapshot_users()` in `includes/network-functions.php` runs one `get_users( include, fields => ID )` over every user online anywhere on the network on every network-admin request, purely so the counts are right during that window. Measured on a seeded network of 1000 sites x 15 users:

| | cold read |
|---|---|
| with the existence query | 39.0 ms |
| existence query short-circuited | 19.5 ms |

Half the read path exists to cover a hook that was never added.

Fix: hook `deleted_user` (and `remove_user_from_blog`) to delete the user's presence rows and re-push the affected sites, then reconsider whether the read-path filter is still worth its cost.

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

</details>

josephfusco:

1. #332 summary table
2. **#333 write path (this one)**
3. #334 read path
4. #335 Sites list column
5. #336 Users list

---

The table in #332 has nothing writing into it.

Part of #298
Closes #327

<details>

AI assistance: Yes
Tool(s): Claude Code

</details>

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -116,6 +116,10 @@ function wp_set_presence( $room, $client_id, $state, $user_id = 0 ) {
 		)
 	);
 
+	if ( false !== $result && wp_presence_admin_room() === $room ) {
+		wp_presence_admin_room_changed();
+	}
+
 	return false !== $result;
 }
 
@@ -143,6 +147,10 @@ function wp_remove_presence( $room, $client_id ) {
 		array( '%s', '%s' )
 	);
 
+	if ( false !== $result && wp_presence_admin_room() === $room ) {
+		wp_presence_admin_room_changed();
+	}
+
 	return false !== $result;
 }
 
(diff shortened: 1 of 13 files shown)


josephfusco:
Review findings from #299 that land here are #310, answered by the debounce, plus #329 and #343, which are not fixed.

Issue #328, opened 2026-08-24
Network site links use the current request's scheme for every site


josephfusco:
`includes/network-functions.php:833` builds every site's URL from the scheme of the request being served:

```php
'url' => ( is_ssl() ? 'https://' : 'http://' ) . $site->domain . $site->path,
```

`is_ssl()` describes the current request, not the site being linked to. On a network where sites differ (a subsite still on http, or a mapped domain with its own certificate), the Sites column and the network dashboard widget link to the wrong scheme for every site that does not match whichever one the admin happens to be on.

The comment above the line explains why `get_site_url()` is not used here: it switches blogs on every call, which is the cost this table exists to avoid. That reasoning holds, so the fix is to carry the scheme rather than to start switching. Each site already knows it; `wp_presence_encode_network_summary_row()` writes `$site->domain . $site->path` into the row from inside that site, so it could write the scheme too.

<details>
<summary>Use of AI Tools</summary>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5
Used for: Assisting with analysis

</details>

i-am-chitti:
@josephfusco I would like to pick this too in case you have not started.

josephfusco:
Yours @i-am-chitti, assigned.

Same base note as #327: `includes/network-functions.php` is not on `main` yet. The read side is in #334, the encode side in #333. Wait for both and branch off `main`, or branch off `feature/network-summary-read` now and target that branch.

One thing to plan for: rows written before the change have no scheme, so the read path needs a fallback. The refresh interval in #333 rewrites every row within one TTL, so it can be temporary.

PR #334, opened 2026-08-24
feat: read the network summary as a capped snapshot

### LINKED ISSUE #321: Network summary is rebuilt once per list table row

### LINKED ISSUE #324: Network summary read path is unbounded

josephfusco:

1. #332 summary table
2. #333 write path
3. **#334 read path (this one)**
4. #335 Sites list column
5. #336 Users list

---

The rows #333 writes have nothing reading them, and #310 asked whether that read holds up at large-network scale.

Part of #298
Closes #321
Closes #324

<details>

AI assistance: Yes
Tool(s): Claude Code

</details>

### DIFF
--- a/includes/network-functions.php
+++ b/includes/network-functions.php
@@ -16,11 +16,24 @@
  * same way core registers blogs and sitemeta, so it is created once per
  * network rather than once per site.
  *
+ * Reading splits in two. The snapshot is one query returning user IDs and
+ * nothing else, and it is what most callers want. Resolving those IDs into
+ * display names and avatar URLs is the part that costs, and no surface draws
+ * every user online, so callers ask the summary for the slice they are about
+ * to render rather than taking the whole network and cutting it afterwards.
+ *
  * @package Presence_API
  */
 
 if ( ! defined( 'ABSPATH' ) ) {
 	exit;
+}
+
+// How many people a network surface draws per site. Every one of them costs a
+// user load and an avatar URL, so this is also what the read path is asked to
+// resolve; the headcount beside the stack is the site's real total.
+if ( ! defined( 'WP_PRESENCE_NETWORK_AVATARS' ) ) {
+	define( 'WP_PRESENCE_NETWORK_AVATARS', 4 );
 }
 
 /**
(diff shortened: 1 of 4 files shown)


josephfusco:
Review findings from #299 and #313 that land here are #321 and #324, both closed by this one, with #327, #328, and #344 still open against the same code.

PR #336, opened 2026-08-24
feat: add an Online view and column to the Network Users list

josephfusco:
Part of the network presence stack, merged bottom-up:

1. #332 summary table
2. #333 write path
3. #334 read path
4. #335 Sites list column
5. **#336 Users list (this one)**
6. #337 network dashboard widget

---

The summary is keyed by site and this screen is keyed by user.

Part of #298
Closes #322

<details>
<summary>Use of AI Tools</summary>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Sonnet 5, Claude Opus 5
Used for: Assisting with design, implementation, and testing

</details>

Issue #331, opened 2026-08-24
Avatar stack renders an empty img server-side where the widget JS renders nothing


josephfusco:
The network widget's inline `buildAvatarStack()` skips a user with no `avatar_url`; the server-side `wp_presence_render_avatar_stack()` does not.

`includes/widgets/class-wp-presence-network-widget-whos-online.php`:

```js
if (user.avatar_url) {
    html += '<img src="' + esc(user.avatar_url) + ...
}
```

`includes/functions.php`:

```php
foreach ( array_slice( $users, 0, $stack_max ) as $index => $user ) {
    $html .= '<img src="' . esc_url( $user['avatar_url'] ) . ...
}
```

`get_avatar_url()` returns false when the Show Avatars setting is off or a filter suppresses it, so on such a network the first render paints a broken image per user and the first heartbeat tick replaces it with nothing. The two renders have to agree; that is the whole point of the shared renderer.

Fix: skip the empty `avatar_url` in `wp_presence_render_avatar_stack()`, matching the JS. The z-index then has to come off the rendered index rather than the sliced one.

Reported by @josephfusco in the review on #313.

---

**Where:** `wp_presence_render_avatar_stack()` in `includes/functions.php`, introduced by #335 and not on `main` yet. Either branch off `feature/network-sites-column` and target that branch, or wait for #335 to land and branch off `main`.

**Test:** `npx wp-env start` then `npm test`. Avatar stack coverage lives in `tests/test-functions.php`; a regression wants a user whose `avatar_url` is empty and an assertion that no `<img>` is rendered for them.

PR #337, opened 2026-08-24
feat: add a Who's Online widget to the Network Admin dashboard

### LINKED ISSUE #325: Network widget heartbeat sends the full summary whenever anything on the network changes

josephfusco:

1. #332 summary table
2. #333 write path
3. #334 read path
4. #335 Sites list column
5. #336 Users list
6. **#337 network dashboard widget (this one)**

---

Last surface, and the only one that ticks on a heartbeat rather than rendering once.

Part of #298
Closes #325

<details>

AI assistance: Yes
Tool(s): Claude Code

</details>

### DIFF
--- a/includes/widgets/class-wp-presence-network-widget-whos-online.php
+++ b/includes/widgets/class-wp-presence-network-widget-whos-online.php
@@ -1,0 +1,330 @@
+<?php
+/**
+ * Network Dashboard Widget: Who's Online
+ *
+ * @package Presence_API
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Handles the network dashboard's "Who's Online" widget with Heartbeat integration.
+ *
+ * The one new surface this plugin adds for multisite: everything else folds
+ * into existing Network Admin screens (Sites and Users list columns), but a
+ * dashboard widget mirrors the single-site "at a glance" pattern and has no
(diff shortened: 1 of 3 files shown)


josephfusco:
Review findings from #313 that land here are #325, closed by this one, and #331, which is not.

PR #339, opened 2026-08-24
fix: clear a user's presence when their account or site membership ends

### LINKED ISSUE #327: Deleting a user leaves their presence behind on every site they were online on

i-am-chitti:
### Description

Fixes #327

Hooks `deleted_user` and `remove_user_from_blog` to clear a user's presence immediately instead of leaving rows to age out on TTL. Re-pushing the network summary falls out for free — `wp_remove_user_presence()` already fires the change signal the push is wired to.

### Testing
- Added unit tests for both hooks in isolation, plus an integration test covering `wpmu_delete_user()` clearing presence across every site a user belonged to.
- Updated two existing tests (post list, REST controller) that had relied on a deleted user's row surviving deletion to exercise their own fallback-rendering paths.

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Sonnet 5
Used for: Implementation and tests

### DIFF
--- a/includes/lifecycle.php
+++ b/includes/lifecycle.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Lifecycle hooks: sets/removes presence on login and logout.
+ * Lifecycle hooks: sets/removes presence on login, logout, and user removal.
  *
  * @package Presence_API
  */
@@ -39,4 +39,18 @@ function wp_presence_on_logout( $user_id ) {
 	if ( $user_id && user_can( $user_id, 'edit_posts' ) ) {
 		wp_remove_user_presence( $user_id );
 	}
+}
+
+/**
+ * Removes a user's presence when their account is deleted or they are
+ * removed from a site.
+ *
+ * Hooked to 'deleted_user' and 'remove_user_from_blog', both of which fire
+ * with the site already switched to the one the user is leaving. No
+ * capability gate, unlike login/logout: their role may already be gone.
+ *
+ * @param int $user_id The ID of the user being deleted or removed.
+ */
+function wp_presence_on_user_removed( $user_id ) {
+	wp_remove_user_presence( $user_id );
 }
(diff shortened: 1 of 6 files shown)


josephfusco:
Base branch is gone. #299 and #313 were closed today and replaced by a stack, #332 through #337, so `feature/network-presence-screen` will never merge.

Retarget to `main`. Your commits cherry-pick clean apart from one test in `tests/test-network-presence.php`, which the restructure split into four files. That test asserts the re-push, so it belongs in `tests/test-network-summary-push.php` on #333, where it passes unchanged.

#328 has the same base problem; note on the issue.

josephfusco (approved):
This looks good to me. Thanks for adding!

PR #356, opened 2026-08-25
fix: skip notifications for no-op presence changes

### LINKED ISSUE #343: Presence removal signals a change when the delete matched no rows
`wp_remove_user_presence()` guards on `false !== $result`, but `$wpdb->delete()` returns `0` when nothing matched, and `false !== 0` is true.

`includes/functions.php:178`

```php
if ( false !== $result ) {
	wp_presence_admin_room_changed();
}
```

So a pagehide or logout for a user with no presence rows still fires `wp_presence_admin_room_changed()`, and the network push runs its compare SELECT against `{base_prefix}presence_network_summary` before deciding not to write. Same shape at `includes/functions.php:119` and `:150`.

Guard on `$result > 0` instead.

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

</details>

---

**Where:** `includes/functions.php:178`, and the same shape at `:119` and `:150`. All three are on `main`, so branch off `main`.

**Test:** `npx wp-env start` then `npm test`. `tests/test-functions.php` is the home for it. The regression wants a removal for a user with no presence rows and an assertion that `wp_presence_admin_room_changed` does not fire.

mori-ikuri:
## Summary

- fire `wp_presence_admin_room_changed()` only when a presence write or delete changes at least one row
- preserve the existing successful return value for zero-row no-op operations
- add regression coverage for no-op client and user presence removal

## Testing

- `php -l includes/functions.php`
- `php -l tests/test-functions.php`
- `vendor/bin/phpcs --standard=phpcs.xml.dist includes/functions.php`
- `vendor/bin/phpstan analyse --configuration=phpstan.neon.dist --memory-limit=2G --no-progress`
- focused PHP harness covering database results `0`, `>0`, and `false`

Not run locally: the wp-env PHPUnit suite because Docker is unavailable in the local environment.

Fixes #343

<details>

AI assistance: Yes
Tool(s): ChatGPT, OpenAI Codex
Model(s): GPT-5
Used for: Candidate research, implementation, regression-test drafting, local validation, diff review, and PR text preparation. The Owner reviewed the change before submission.

</details>

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -116,7 +116,7 @@ function wp_set_presence( $room, $client_id, $state, $user_id = 0 ) {
 		)
 	);
 
-	if ( false !== $result && wp_presence_admin_room() === $room ) {
+	if ( $result > 0 && wp_presence_admin_room() === $room ) {
 		wp_presence_admin_room_changed();
 	}
 
@@ -147,7 +147,7 @@ function wp_remove_presence( $room, $client_id ) {
 		array( '%s', '%s' )
 	);
 
-	if ( false !== $result && wp_presence_admin_room() === $room ) {
+	if ( $result > 0 && wp_presence_admin_room() === $room ) {
 		wp_presence_admin_room_changed();
 	}
 
@@ -175,7 +175,7 @@ function wp_remove_user_presence( $user_id ) {
 	);
 
 	// Deletes across every room, so the admin room is always among them.
-	if ( false !== $result ) {
+	if ( $result > 0 ) {
 		wp_presence_admin_room_changed();
 	}
 
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -195,8 +195,7 @@ function wp_presence_admin_room_changed() {
 	/**
 	 * Fires after a write that may have changed who's online on this site.
 	 *
-	 * Fires on every admin-room write, including a tick that refreshes the same
-	 * people; listeners that only want real changes compare state themselves.
+	 * Fires when an admin-room write changes at least one row.
 	 *
 	 * @since 0.1.25
 	 */
--- a/tests/test-functions.php
+++ b/tests/test-functions.php
@@ -118,6 +118,25 @@ class WP_Test_Presence_Functions extends WP_Presence_UnitTestCase {
 	}
 
 	/**
+	 * @covers ::wp_remove_presence
+	 * @covers ::wp_presence_admin_room_changed
+	 */
+	public function test_remove_nonexistent_admin_presence_does_not_announce_a_change() {
+		$fired = 0;
+		add_action(
+			'wp_presence_admin_room_changed',
+			function () use ( &$fired ) {
+				++$fired;
+			}
+		);
+
+		$result = wp_remove_presence( wp_presence_admin_room(), 'nonexistent' );
+
+		$this->assertTrue( $result, 'A no-op removal should still report success.' );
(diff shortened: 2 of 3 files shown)


i-am-chitti:
Hey @mori-ikuri thanks for tackling this.

> The following contributors have not linked their GitHub and WordPress.org accounts: @mori-ikuri

To receive proper attribution please associate a WordPress.org account with your GitHub account. You can read about it [here](https://make.wordpress.org/core/2020/03/19/associating-github-accounts-with-wordpress-org-profiles/). Let us know once that's setup and we'll run the props-bot again.

mori-ikuri:
@i-am-chitti 
Thanks! I've linked my WordPress.org account (moriikuri) with my GitHub account. Could you please rerun props-bot?

i-am-chitti:
@mori-ikuri You've been referenced here - https://github.com/WordPress/presence-api/pull/356#issuecomment-5406219118

includes/functions.php
```diff
@@ -116,7 +116,7 @@ function wp_set_presence( $room, $client_id, $state, $user_id = 0 ) {
 		)
 	);
 
-	if ( false !== $result && wp_presence_admin_room() === $room ) {
+	if ( $result > 0 && wp_presence_admin_room() === $room ) {
```
i-am-chitti, line 119:
```
WP_Test_Network_Summary_Push
 ✘ Push refreshes a stale row when nobody changed
   ┐
   ├ Failed asserting that '2026-08-25 15:42:20' is greater than '2026-08-25 15:42:20'.
   │
   ╵ /var/www/html/wp-content/plugins/presence-api/tests/test-network-summary-push.php:143
   ┴

FAILURES!
Tests: 351, Assertions: 780, Failures: 1.
```

One test is failing. I think the guard is fine and the test is the problem. It writes presence twice back to back in the same second, so the upsert affects 0 rows and the hook doesn't fire. A real heartbeat tick is always seconds apart, so that path won't happen in reality.

Aging the presence row 2 seconds before the second write makes it pass with `$result > 0`. @josephfusco How should we handle this? Should we skip the guard here or fix the test?

One more thing to call: [wp_presence_admin_room_changed](https://github.com/WordPress/presence-api/blob/c41076c4cb14e8269316c0f98eb197349e4cefb2/includes/functions.php#L198-L199) is documented as *"Fires on every admin-room write, including a tick that refreshes the same people."* This change narrows it slightly, so the docblock may need an update.

i-am-chitti (commented):
@josephfusco the multisite test runs only in CI. Nothing in `package.json` covers it, and plain `npm test` skips tests silently. For example, on this PR. Given how much network functionality is multisite-only, I propose that a dedicated `test:multisite` script would be worth adding. What do you think?

josephfusco:
@i-am-chitti Filing a separate issue for the npm scripts gap.

Tests need to pass before this merges; other feedback can follow up separately. Thanks for the contribution, @mori-ikuri.

includes/functions.php
josephfusco: Fix the test: age the presence row 2 seconds before the second `set_presence_on_site` call so the upsert produces a changed `date_gmt` and `$result > 0`. Real heartbeat ticks are always seconds apart, so this matches production behavior. The guard is correct.

Docblock should drop the "including a tick that refreshes the same people" clause -- tighten to "Fires when an admin-room write changes at least one row."

Pushing both fixes to the branch now.

PR #357, opened 2026-08-25
perf: gate network presence aggregation on wp_is_large_network()

### LINKED ISSUE #348: Network summary push has no large-network gate and no way to turn it off
`wp_presence_push_network_summary()` guards on two things and neither of them is scale.

`includes/network-functions.php:176`

```php
function wp_presence_push_network_summary() {
	if ( ! is_multisite() || ! wp_presence_has_network_summary_table() ) {
		return;
	}
```

`wp_is_large_network()` appears twice in the plugin, `presence-api.php:177` and `presence-api.php:254`, both of them site loops during activation and deactivation. It never reaches the write path.

The coalescing in #333 bounds how often one site rewrites its row. It does not bound how many sites write into one table. `presence_network_summary` registers with `base_prefix` into `$wpdb->ms_global_tables`, which is the set a sharded network pins to the global cluster, so the write concentration #310 asks about survives the debounce intact.

There is also no filter to turn aggregation off. `wp_presence_network_summary_refresh_interval` is the only knob and it is clamped at `includes/network-functions.php:147` so a filter cannot widen it. A network above the `wp_is_large_network()` threshold has no way to opt out short of deactivating the plugin.

Fix: a filterable gate defaulting off above `wp_is_large_network()`, checked in `wp_presence_push_network_summary()`. The read path needs the same check, or the network screens show a stale snapshot with no indication that the feed behind it stopped.

Part of #310

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

</details>

i-am-chitti:
### Description

Closes #348

Adds `wp_presence_network_aggregation_enabled()`, defaulting to `! wp_is_large_network()` and filterable both ways. Checked on the push and on the read - a network that stopped aggregating but kept serving its existing rows would draw a snapshot frozen at the last push, with nothing to say the feed had stopped. The read gate sits in `wp_presence_compute_network_snapshot()`, the chokepoint every reader funnels through.

Left open deliberately: a gated network reads as empty, so the screens say nobody is online rather than saying aggregation is off. Happy to add a distinct state if you want one.

### Testing

Both gates are asserted as statement counts, not just absent rows: declining after a `SELECT` against the global cluster still costs what the gate exists to avoid.

Verified on a 3-site network with `wp_is_large_network` forced true. The widget shows its empty state, the Sites column shows `—`, and the Users view shows `Online (0)`.

### Screenshots

dashboard-on

dashboard-off

sites-on

sites-off

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5
Used for: Assisting with the implementation and tests

</details>

### DIFF
--- a/includes/network-functions.php
+++ b/includes/network-functions.php
@@ -70,6 +70,39 @@ function wp_presence_has_network_summary_table() {
 }
 
 /**
+ * Whether this network aggregates presence into the summary table.
+ *
+ * An ms_global_tables entry pins the summary table to the global cluster on a
+ * sharded network, so write concentration grows with the site count however
+ * rarely any single site pushes. Defaults off above wp_is_large_network()
+ * rather than having a network inherit that cost from installing the plugin.
+ *
+ * The read path checks this too, or a network that stopped aggregating would go
+ * on drawing the rows it already had with nothing to say the feed had stopped.
+ *
+ * @access private
+ * @return bool
+ */
+function wp_presence_network_aggregation_enabled() {
+	// Load-bearing rather than redundant: wp_is_large_network() is undefined
+	// outside multisite, and callers here do not check first.
+	if ( ! is_multisite() ) {
+		return false;
+	}
+
+	/**
+	 * Filters whether network-wide presence aggregation runs.
+	 *
+	 * Gates the push and the reads behind the Network Admin screens.
+	 *
+	 * @since 0.2.0
+	 *
+	 * @param bool $enabled Default false above wp_is_large_network().
+	 */
+	return (bool) apply_filters( 'wp_presence_network_aggregation_enabled', ! wp_is_large_network() );
+}
+
+/**
  * Creates or updates the network-wide presence summary table if needed.
  *
  * One table for the whole network rather than one per site, since it exists
@@ -174,7 +207,7 @@ function wp_presence_network_summary_refresh_interval() {
  * @access private
  */
 function wp_presence_push_network_summary() {
-	if ( ! is_multisite() || ! wp_presence_has_network_summary_table() ) {
+	if ( ! wp_presence_network_aggregation_enabled() || ! wp_presence_has_network_summary_table() ) {
 		return;
 	}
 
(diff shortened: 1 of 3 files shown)


i-am-chitti:
Actions are failing WordPress org-wide due to some billing issue.

includes/network-functions.php
```diff
@@ -69,6 +69,39 @@ function wp_presence_has_network_summary_table() {
...
+	/**
+	 * Filters whether network-wide presence aggregation runs.
+	 *
+	 * Gates the push and the reads behind the Network Admin screens.
+	 *
+	 * @since 0.2.0
+	 *
+	 * @param bool $enabled Default false above wp_is_large_network().
+	 */
+	return (bool) apply_filters( 'wp_presence_network_aggregation_enabled', ! wp_is_large_network() );
```
josephfusco, line 102: Is a filter needed here? Core's wp_is_large_network filter already takes $component, so an operator can flip just the sites axis without a plugin-specific hook.

josephfusco (changes requested):

includes/network-functions.php
i-am-chitti: The default here already comes from `wp_is_large_network()`, so the standard behaviour is free.

The catch is that core's filter changes other things too. To turn aggregation on for a big network you would set the sites axis to `false`, which also brings back [paging](https://github.com/WordPress/wordpress-develop/blob/7.1.0/src/wp-admin/includes/class-wp-ms-sites-list-table.php#L97) and the [count query](https://github.com/WordPress/wordpress-develop/blob/7.1.0/src/wp-admin/includes/class-wp-ms-sites-list-table.php#L165) on Network Admin > Sites, and makes this plugin [loop through every site on activation](https://github.com/WordPress/presence-api/blob/920d190f2d104a94fb7cfbc4bec3e746c10d2761/presence-api.php#L177). A loop that is skipped on large networks on purpose. Setting it to `true` on a small network breaks the same things in reverse.

So core's filter answers "is this network big" and this one answers "should presence aggregate". LMK what you think.

josephfusco: That makes sense, thanks for clarifying.  I'm good with the new filter being added to cover that use case.

josephfusco (approved):

PR #365, opened 2026-08-26
feat: expose network presence via REST and WP-CLI

### LINKED ISSUE #350: No REST route or CLI command for network presence
Single-site presence is reachable two ways besides the admin UI:

- `includes/class-wp-rest-presence-controller.php`
- `includes/cli/class-wp-presence-cli-command.php`

Network presence has neither. `wp_presence_get_network_summary()` at `includes/network-functions.php:507` is called only from the three Network Admin screens, so there is no way to ask who is online across a network from a script, a monitoring check, or another plugin, and no way to inspect the summary table without SQL.

Fix: a `wp presence network` CLI subcommand and a network-scoped REST route, both gated on `wp_presence_network_capability()`.

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

</details>

i-am-chitti:
### Description

Closes #350

Adds a `wp presence network` subcommand and two network-scoped REST routes, both gated on `wp_presence_network_capability()`. Both read the summary table directly, so answering for the whole network costs no `switch_to_blog()` per site.

Sites are the paginated unit, since a site is what the table holds a row for. `X-WP-Total` counts sites, and the user headcount rides in `X-WP-Presence-Users-Online` — a caller after the total alone would otherwise have to walk every page for it. Paginating meant giving `wp_presence_get_network_summary()` an `offset`, which is the only change here to existing code.

Two decisions worth a look:

- The CLI checks the capability only once `--user` names someone. Shell access already outranks `manage_network`, so the check keeps `--user` honest rather than keeping anyone out — but the guard reads oddly without that context.
- A site nobody is on answers `200` with an empty user list, so only an unknown `blog_id` is a `404`. A poll against one site reads the same shape either way.

### Testing

Both halves cover the single-site case, where the summary table does not exist: the routes are not registered, and the subcommand refuses. That path is invisible to the multisite suite, and without the guard the subcommand fatals on an undefined function rather than erroring cleanly.

#### WP CLI

image

image

image

image

image

#### REST API

image

image

image

image

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5
Used for: Assisting with the implementation and tests

</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,17 @@ All endpoints require `edit_posts`. Responses include `Cache-Control: no-store`.
 | `DELETE` | `/wp-presence/v1/presence` | Remove a presence entry |
 | `GET` | `/wp-presence/v1/presence/rooms` | List active rooms |
 
+### Network
+
+Multisite only, and gated on `manage_network` rather than `edit_posts`.
+
+| Method | Path | Description |
+|---|---|---|
+| `GET` | `/wp-presence/v1/presence/network` | List sites with users online, busiest first |
+| `GET` | `/wp-presence/v1/presence/network/<blog_id>` | One site's users online |
+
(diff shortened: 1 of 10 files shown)


josephfusco (approved):
Great additions!

PR #366, opened 2026-08-26
Playground preview: derive PR from the triggering run

obenland:
## What

The `playground-preview-publish.yml` workflow reads `pr_number` and `head_sha` from the build's `pr-meta.json`. This change instead:

- Requires the SHA recorded in `pr-meta.json` to match `github.event.workflow_run.head_sha`; otherwise it refuses to publish.
- Resolves the PR number from that trusted head SHA (open PRs matched on `head.sha`), rather than trusting the artifact's `pr_number`.

The prerelease and the sticky Playground comment now target the PR the built commit actually belongs to.

## Why

The build runs PR code under a read-only token, so `pr-meta.json` is untrusted. The workflow already routes Playground at base-repo release URLs only (good), but the PR number it targeted was still artifact-supplied — so the release tag and preview comment could be pointed at another open PR. Binding to the `workflow_run` head SHA closes that.

The producer already installs with `--no-scripts`, so no build-side change is needed.

## Testing

- `actionlint` clean (includes shellcheck on the `run` block).
- Resolution logic exercised locally against synthetic `gh api` responses: honest run resolves its PR; a mismatched SHA is refused; no matching open PR fails closed; duplicate head SHAs resolve without a pipefail edge.
- Worth a check on a fork-originated PR to confirm the PR resolves in the runner.

### DIFF
--- a/.github/workflows/playground-preview-publish.yml
+++ b/.github/workflows/playground-preview-publish.yml
@@ -67,6 +67,12 @@ jobs:
             }
 
       - name: Unpack and validate artifacts
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+          # The build runs untrusted PR code; the run's head SHA is the only
+          # trusted value.
+          TRUSTED_HEAD_SHA: ${{ github.event.workflow_run.head_sha }}
         run: |
           set -euo pipefail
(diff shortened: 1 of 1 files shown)


josephfusco (approved):
Confirmed on a fork PR: #357's build run has `head_sha` `1d4e073dcbe4`, matching its `head.sha`, while `workflow_run.pull_requests` is empty, so the lookup resolves exactly what the payload can't.

PR #371, opened 2026-08-26
fix: bring the stale-screen banner back after a dismissal

### LINKED ISSUE #361: Dismissing the stale-screen banner suppresses it for the life of the page
`bannerShown` is never reset ([`assets/js/stale-screen.js:104`](https://github.com/WordPress/presence-api/blob/1f7bd622d8ec9e31506478a592bf01954079c744/assets/js/stale-screen.js#L104)) and the dismiss handler only calls `notice.remove()` ([`:153`](https://github.com/WordPress/presence-api/blob/1f7bd622d8ec9e31506478a592bf01954079c744/assets/js/stale-screen.js#L153-L155)), so one dismissal swallows every later edit by another user.

jigneshbhavani:
Dismissing the banner removed the notice but left `bannerShown` set, so the guard in the tick handler suppressed every later banner for the life of the page. One dismissal and you stop hearing about other people's edits until you reload.

Clearing the flag alone is not enough. The tick handler compares against `baselineRev`, which the banner path deliberately does not advance, so a dismissal that only resets the flag brings the same banner straight back on the next tick. The dismiss handler now also advances `baselineRev` to the revision that raised the banner, which is what makes a dismissal mean "I have seen this edit" rather than "hide this screen".

One consequence worth knowing: if a newer edit lands while the banner is already up, the flag suppresses it and the baseline only advances to the older revision, so dismissing raises a fresh banner on the next tick. That is intended, there really is an edit you have not seen, but it does mean a dismissal can be followed immediately by another banner.

Three e2e guards in the existing stale-screen file, each mutation checked to fail on its own and only its own:

- the banner returns after a later edit, which fails without the flag reset
- a dismissal sticks when nothing else has happened, which fails without the baseline advance
- your own save stays silent and the next foreign edit still reports. That branch had no coverage at all: deleting it broke no test before this

Two things found while writing those, neither fixed here:

`wp_update_post()` never writes `_edit_last`, because core only does that from the admin save path. An edit driven from `wp eval` is therefore authorless and `actor_is_me` is always false. The relay test at `:160` sets a current user expecting that to attribute the edit, and it does not. My helper sets the meta explicitly. Happy to correct the older test here or leave it for you.

`presence-widgets.test.js` "Focus stays on the post link in Active Posts widget across a heartbeat re-render" is flaky on `main` independently of this branch: 5 of 6 runs failed on `005b6ca` untouched, 2 of 6 with this change. Worth its own issue if it is not already known.

Fixes #361

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

Used for: Diagnosis, the fix, and the e2e guards. I reviewed the change, ran the mutation checks and the suites, and confirmed the flaky test against an untouched base branch.

</details>

### DIFF
--- a/assets/js/stale-screen.js
+++ b/assets/js/stale-screen.js
@@ -93,10 +93,10 @@
 		if ( bannerShown ) {
 			return;
 		}
-		showBanner( info );
+		showBanner( info, rev );
 	} );
 
-	function showBanner( info ) {
+	function showBanner( info, rev ) {
 		const target = document.querySelector( '.wrap' ) || document.getElementById( 'wpbody-content' );
 		if ( ! target ) {
 			return;
@@ -152,6 +152,11 @@
 		dismiss.appendChild( sr );
 		dismiss.addEventListener( 'click', function () {
 			notice.remove();
+			bannerShown = false;
+			// Dismissal acknowledges this revision, not the screen. The tick
+			// handler only compares against baselineRev, so clearing the flag
+			// without advancing it brings the same banner back one tick later.
+			baselineRev = rev;
 		} );
 		notice.appendChild( dismiss );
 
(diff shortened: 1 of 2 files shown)


josephfusco (approved):
This looks good!

PR #373, opened 2026-08-27
refactor: share the avatar stack's CSS and JS across every surface

### LINKED ISSUE #312: Avatar-stack CSS and JS are duplicated between the single-site and network widgets
#299 added a network dashboard widget alongside the existing single-site "Who's Online" widget. Both widgets render an overlapping avatar stack, and #299's follow-up commit shares the PHP rendering (wp_presence_render_avatar_stack() in includes/functions.php), but the CSS and the client-side Heartbeat-update JS are still separate copies.

Each widget class defines its own .presence-avatar-stack CSS block (same properties, different container ID prefix and image size), and its own inline buildAvatarStack() JS function (same z-index logic, different image size: 20px for the network widget, 24px for the single-site widget).

Consolidating these means deciding how the two widgets share a script/style asset rather than each carrying an inline copy generated by wp_add_inline_script()/wp_add_inline_style().

### LINKED ISSUE #331: Avatar stack renders an empty img server-side where the widget JS renders nothing

i-am-chitti:
Shared `assets/css/avatar-stack.css` and `assets/js/avatar-stack.js` replace the four inline copies. Each widget's inline script moves off the `heartbeat` handle onto one of its own, so it can declare the shared file as a dependency. `$size` is gone; both callers passed a value the CSS overrode.

The shared renderer is where the server and Heartbeat renders have to agree, which is #331: `wp_presence_render_avatar_stack()` now skips a user with no `avatar_url` instead of emitting ``, matching what the widget JS already did.

Separately, nothing enqueued a stylesheet on Network Admin > Sites, so the column added in #335 has been drawing square, ringless avatars. It gets the shared stylesheet now.

| Before | After |
| --- | --- |
| sites-column-before | sites-column-after |

Fixes #312
Fixes #331

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5
Used for: Implementation and tests, reviewed and edited by me.

</details>

### DIFF
--- a/assets/css/avatar-stack.css
+++ b/assets/css/avatar-stack.css
@@ -1,0 +1,27 @@
+/**
+ * Overlapping avatar stack.
+ *
+ * Shared by the two "Who's Online" dashboard widgets and the "Online" column
+ * on Network Admin > Sites.
+ *
+ * @package Presence_API
+ */
+
+/* Inline so the stack sits on the baseline beside the Sites list count. */
+.presence-avatar-stack {
+	display: inline-flex;
+	align-items: center;
+}
+
+.presence-avatar-stack img {
+	position: relative;
+	width: 20px;
+	height: 20px;
+	margin-inline-start: -6px;
+	border-radius: 50%;
+	box-shadow: 0 0 0 2px #fff;
+}
+
+.presence-avatar-stack img:first-child {
+	margin-inline-start: 0;
+}
--- a/assets/js/avatar-stack.js
+++ b/assets/js/avatar-stack.js
@@ -1,0 +1,41 @@
+/**
+ * Overlapping avatar stack, client side.
+ *
+ * The counterpart to wp_presence_render_avatar_stack(). Both "Who's Online"
+ * widgets paint the stack server-side and then repaint it from a Heartbeat
+ * tick, so the two renderers have to emit the same markup.
+ *
+ * @package Presence_API
+ */
+( function () {
+	'use strict';
+
+	function esc( str ) {
+		var el = document.createElement( 'span' );
+		el.textContent = str;
+		return el.innerHTML;
+	}
+
+	/**
+	 * Builds an avatar stack.
+	 *
+	 * @param {Array}  users Users, each with an avatar_url and a display_name.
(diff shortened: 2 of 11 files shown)


josephfusco:
That last commit is on the bot, not you: `wpPresenceBuildAvatarStack` is only called from the widgets' own inline scripts, so it wants `@private` on the docblock rather than a README entry, and I will ping you once the check is fixed to accept that.

josephfusco:
That is merged, so revert 171b4a1 and put `@private` on the docblock in `assets/js/avatar-stack.js` instead.

i-am-chitti:
@josephfusco Thanks. I've marked it private. Ready for a review now.

josephfusco (approved):
🙌🏼

PR #374, opened 2026-08-27
fix: exclude Codecov config and Jest test from release zip

### LINKED ISSUE #370: Release zip ships codecov.yml and a Jest test file
[`.distignore`](https://github.com/WordPress/presence-api/blob/005b6cad3d82c3fcde1867893cdb781f321ce092/.distignore) never picked up `codecov.yml` or `src/utils/test/`, so both ship in the plugin zip and on WordPress.org.

stfulldev:
Excludes `codecov.yml` and `src/utils/test/` from the distributed plugin ZIP.

Fixes #370

## testing

- Built the distribution using the release workflows `rsync` process.
- Confirmed `codecov.yml` is absent.
- Confirmed `src/utils/test/` is absent.
- Confirmed production files under `src/utils/` and `src/hooks/` remain included.

### DIFF
--- a/.distignore
+++ b/.distignore
@@ -18,6 +18,7 @@ blueprint-40.json
 CHANGELOG.md
 README.md
 SECURITY.md
+codecov.yml
 composer.json
 package.json
 package-lock.json
@@ -27,5 +28,6 @@ phpunit.xml.dist
 release-please-config.json
 scripts/
 src/hooks/test/
+src/utils/test/
 includes/db-viewer.php
 includes/debugger-widget.php


i-am-chitti (commented):
LGTM 🚀.  I checked this locally: building with `main`'s `.distignore` does put `codecov.yml` and `src/utils/test` in the zip, and with this patch both are gone. Nothing else in the zip changes.

The failing `e2e / Playwright` check is not caused by this PR. `e2e.yml` doesn't read `.distignore` at all. It's a network flake in `wp-env`, and it takes the whole job down because `retries` is `0` by default. I'll open a separate issue with the details.

i-am-chitti:
@stfulldev Props attribution is missing -

> The following contributors have not linked their GitHub and WordPress.org accounts: @stfulldev

To receive proper attribution please associate a WordPress.org account with your GitHub account. You can read about it [here](https://make.wordpress.org/core/2020/03/19/associating-github-accounts-with-wordpress-org-profiles/). Once you do so, props bot will identify attribution automatically.

josephfusco (approved):
Thank you!

PR #379, opened 2026-08-27
fix: outlast core's unfocused heartbeat interval in the presence TTL

### LINKED ISSUE #354: Presence expires between heartbeats on unfocused and idle tabs
Core drops the heartbeat to 120s on an unfocused or five-minute-idle tab (`scheduleNextTick()` in `wp-includes/js/heartbeat.js`), twice `WP_PRESENCE_DEFAULT_TTL`, so a tab that is still open and still pinging shows offline for 60 seconds of every 120 until core suspends at ten minutes.

Raising the TTL is the only lever: `scheduleNextTick()` ignores `mainInterval` while unfocused, so `wp.heartbeat.interval()` cannot correct it.

Blocks #131.

i-am-chitti:
## Description 

Fixes #354

Raises the default TTL to 150 - the number core sizes post locks against, for the same reason.

Verified in a browser: a visible-but-unfocused tab (`document.hasFocus()` stubbed false), sampled every 10s for 263s.

| elapsed | entry age | |
| --- | --- | --- |
| 61s | 2s | heartbeat tick |
| 122s | 63s | past the old 60s TTL |
| 172s | 114s | peak age, still well inside 150 |
| 182s | 4s | next tick, 120s after the previous one |

Present at all 27 samples. Under the old TTL the tab would have read offline for 60 seconds of every 120, since anything older than 60s fell outside the cutoff. `wp.heartbeat.interval()` reported 60 throughout while the measured gap was 120, which is why the TTL is the only lever. Reverting the constant to 60 makes the new test the only failure in the suite.

Two consequences needs weighing:

- `heartbeat-send` skips pinging while `visibilityState === 'hidden'` so backgrounded tabs age out via the TTL. They now linger up to 150s instead of 60s. Deleting on `visibilitychange` instead is worse: `client_id` is `user-<id>`, shared across tabs, so a hidden tab would clear presence for a user who still has a visible one open.
- `IDLE_THRESHOLD` is unchanged at 30s, so an unfocused tab renders as idle with a "2 mins ago" label between ticks rather than disappearing.

`wp_presence_network_summary_refresh_interval()` needed no change: its `TTL - idleInterval` clamp becomes 105s, and the 120s tick gap still lands inside the 150s cutoff - covered by the `WP_MULTISITE=1` run, which the single-site run skips. The idle backoff still resolves to `min(45, 135) = 45`.

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ System-wide presence and awareness for WordPress.
 
 ## Problem
 
-WordPress has no way to know who is logged in, what screen they are on, or which posts are being edited — without writing to shared tables like `wp_postmeta` or `wp_options`. High-frequency writes to those tables invalidate caches site-wide ([#64696](https://core.trac.wordpress.org/ticket/64696)). This plugin uses a dedicated `wp_presence` table with a 60-second TTL to provide that awareness with zero cache side effects.
+WordPress has no way to know who is logged in, what screen they are on, or which posts are being edited — without writing to shared tables like `wp_postmeta` or `wp_options`. High-frequency writes to those tables invalidate caches site-wide ([#64696](https://core.trac.wordpress.org/ticket/64696)). This plugin uses a dedicated `wp_presence` table with a 150-second TTL to provide that awareness with zero cache side effects.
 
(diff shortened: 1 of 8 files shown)


josephfusco (approved):
🙌🏼

PR #380, opened 2026-08-27
fix: detect the collaboration edge across requests

### LINKED ISSUE #362: Collaboration hooks fire on every tick instead of on the edge
[`includes/heartbeat.php:361`](https://github.com/WordPress/presence-api/blob/1f7bd622d8ec9e31506478a592bf01954079c744/includes/heartbeat.php#L361) holds the previous editor count in a request-scoped `static`, so `$prev_count` is always its `?? 1` default, firing `wp_presence_collaboration_started` on every tick with 2+ editors and leaving `wp_presence_collaboration_ended` unreachable.

jigneshbhavani:
`wp_presence_check_collaboration_threshold()` held the previous editor count in a function `static`. A tick is its own request, so `$prev_count` was always its `?? 1` default: `wp_presence_collaboration_started` fired on every tick with two or more editors, and `wp_presence_collaboration_ended` could not fire at all.

Measured with two logged-in sessions on the same post, both hooks logged. On `2acf1fe`, six ticks from the second editor produced `started=7` (the sixth plus one automatic tick from the other session) and `ended=0`. The same run on this branch gives `started=1, ended=1`.

**Why the suite was green on both hooks.** `test_collaboration_started_action` and `test_collaboration_ended_action` call the handler several times inside one PHPUnit process, which is the one condition where a `static` does work. They cover the sequence without the request boundary that breaks it, and they still pass if you revert this change. The four new tests seed the stored count the way an earlier request would have left it, which is what makes them fail on the old code.

**The decision you did not specify: a per-room transient, not an option.** An option grows a row per post anyone edits and never reclaims it. An expiring transient is not autoloaded, and tying its TTL to `wp_presence_get_timeout( WP_PRESENCE_DEFAULT_TTL )` means the memory of a room dies exactly when the entries it describes do. That expression also means #379 raising the TTL to 150 carries this along with no edit here.

**It rewrites on every tick rather than only when the count moves,** which is one row per active room per tick. Writing only on change was my first version and it is wrong: the expiry never moves, so a steady pair lapses mid-session and announces itself again. `test_collaboration_state_expiry_is_pushed_forward_on_every_tick` is the guard for exactly that. If that write volume is the wrong trade, say so, because the alternative is a longer fixed TTL and I would rather you picked it.

**Known limitation.** Read-then-write is not atomic, so two editors ticking in the same instant can both see the old count and both fire `started`. Rarer than the current behaviour of firing every tick, but not zero.

Note #379 also edits `tests/test-heartbeat.php`, so whichever of us lands second gets a trivial rebase.

Fixes #362

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

Used for: Diagnosis, the fix, and the tests. I reproduced the every-tick firing on a running site with two logged-in sessions before any code was written, verified the fix against the base branch the same way, and ran the mutation checks and suites myself.

</details>

### DIFF
--- a/includes/heartbeat.php
+++ b/includes/heartbeat.php
@@ -337,6 +337,22 @@ function wp_presence_editor_heartbeat_received( $response, $data, $screen_id ) {
 }
 
 /**
+ * Builds the transient key holding a room's last observed editor count.
+ *
+ * @since 0.2.0
+ *
+ * @access private
+ * @param string $room The presence room identifier.
+ * @return string The transient key.
+ */
+function wp_presence_collaboration_state_key( $room ) {
+	// Rooms are `postType/post:123` and similar, so they carry characters an
+	// option name may not, and they are longer than the 172 a transient key
+	// can hold once the `_transient_timeout_` prefix is added.
+	return 'wp_presence_collab_' . md5( $room );
+}
+
+/**
  * Checks if the collaboration threshold has been crossed and fires appropriate actions.
  *
  * Fires 'wp_presence_collaboration_started' when editor count goes from 1 to 2+.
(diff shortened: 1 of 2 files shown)


josephfusco (approved):
🙌🏼

PR #392, opened 2026-08-28
fix: Filter out archived, spam, and deleted sites from network presence

### LINKED ISSUE #12: Alert when a screen is stale due to a user having updated settings
I don't think this is currently accounted for:

1. User A navigates to a settings page and starts making changes.
2. User B sees that User A is on that page and navigates there. 
3. User B sees the already-saved settings.
4. User A saves the settings. For classic settings screens, this is a redirect back to the screen.
5. User B never sees any changes to the settings until they explicitly reload.

It seems there should be some indicator for User B that the screen they're looking at is stale. For classic settings pages detecting a save is easier since a full page redirect happens. For custom settings screens or screens which rely entirely on JavaScript to persist the settings, there wouldn't be a standard way to hook into detecting when the screen goes stale.

### LINKED ISSUE #351: Network presence lists archived and spam sites as live
Three `get_sites()` calls in `includes/network-functions.php` pass only `site__in`:

- `:428`
- `:714`
- `:808`

`WP_Site_Query` defaults `archived`, `spam`, `mature`, and `deleted` to `''`, which means no filter. An archived or spam site with a fresh summary row is therefore listed next to live ones, and on the Network Admin dashboard widget it reads as a working site with people on it. `WP_MS_Sites_List_Table` marks those statuses; the widget has nowhere to.

Fix: decide per surface rather than globally. The Sites list column should match whatever the list table it sits in is already filtered to. The dashboard widget and the Users list should exclude archived, spam, and deleted.

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

</details>

---

**Where:** three `get_sites()` calls in `includes/network-functions.php` at `:428`, `:714`, and `:808`, introduced across #334 and above and not on `main` yet. Wait for those to land and branch off `main`.

**Test:** multisite only. `npx wp-env start` then `WP_MULTISITE=1 npm test`. `tests/test-network-summary-push.php` and `tests/test-network-sites-column.php` are the closest homes; the regression wants an archived site with a fresh summary row.

ishitaj34:
`get_sites()` was previously called with only `site__in`, and because `WP_Site_Query` defaults `archived`, `spam`, and `deleted` parameters to empty strings (`''`), no status filtering was performed. Archived or spammed sites with fresh summary rows were returned and treated as live across network surfaces.

**Changes Made:**
- Added explicit status exclusions (`'archived' => 0, 'spam' => 0, 'deleted' => 0`) to three `get_sites()` calls in `includes/network-functions.php`.
  - `wp_presence_compute_network_snapshot()`: Excludes dead sites from the network snapshot powering the Network Admin Dashboard Widget and Users list view count.
  - `wp_presence_hydrate_network_snapshot()`: Guards site object hydration during rendering.
  - `wp_presence_get_network_sites_for_user()`: Ensures the Network Users list column omits archived/spam/deleted sites.
- Added `$sites = is_array( $sites ) ? $sites : array();` to `uninstall.php` for array type safety.
- Added regression tests covering each surface.
  - `test_snapshot_excludes_deleted_site_with_fresh_row` in `tests/test-network-summary-push.php` (tests `deleted` status flag)
  - `test_sites_column_shows_dash_for_archived_site` in `tests/test-network-sites-column.php` (tests `archived` status flag)
  - `test_users_column_omits_spam_site` in `tests/test-network-users-list.php` (tests `spam` status flag)

Fixes #351

<details>

- AI assistance: Yes
- Tool(s): Claude, Gemini
- Model(s): 3.7 Sonnet, 3.6 Flash
- Used for: Understanding the codebase architecture, researching default query behaviors, reviewing the per-surface fix, drafting unit tests, and adding inline documentation; final implementation, code formatting, and test execution were reviewed and verified by contributor.

</details>

### DIFF
--- a/presence-api.php
+++ b/presence-api.php
@@ -217,13 +217,15 @@ function wp_presence_on_initialize_site( $site ) {
  * @return int[] Site IDs.
  */
 function wp_presence_get_network_site_ids() {
-	return get_sites(
+	$sites = get_sites(
 		array(
 			'fields'                 => 'ids',
 			'number'                 => 0,
 			'update_site_meta_cache' => false,
 		)
 	);
+
+	return is_array( $sites ) ? $sites : array();
 }
 
 /**
--- a/includes/network-functions.php
+++ b/includes/network-functions.php
@@ -425,13 +425,18 @@ function wp_presence_get_network_sites_for_user( $user_id ) {
 	// Resolved here rather than in the index above, so a network where nobody
 	// is looked up pays for no sites at all. get_sites() primes the site cache,
 	// so rows sharing a site resolve it once.
-	return get_sites(
+	$sites = get_sites(
 		array(
 			'site__in' => $blog_ids,
 			'number'   => 0,
 			'orderby'  => 'site__in',
+			'archived' => 0,
+			'spam'     => 0,
+			'deleted'  => 0,
 		)
 	);
+
+	return is_array( $sites ) ? $sites : array();
 }
 
 /**
@@ -716,8 +721,15 @@ function wp_presence_compute_network_snapshot( $timeout ) {
 			'site__in' => array_keys( $by_site ),
 			'fields'   => 'ids',
 			'number'   => 0,
+			'archived' => 0,
+			'spam'     => 0,
+			'deleted'  => 0,
 		)
 	);
+
+	if ( ! is_array( $live ) ) {
+		$live = array();
+	}
 
 	$by_site = array_intersect_key( $by_site, array_flip( array_map( 'intval', $live ) ) );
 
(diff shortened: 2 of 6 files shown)


presence-api.php
```diff
@@ -229,13 +217,15 @@ function wp_presence_on_initialize_site( $site ) {
...
  */
 function wp_presence_get_network_site_ids() {
-	return get_sites(
+	$sites = get_sites(
 		array(
 			'fields'                 => 'ids',
 			'number'                 => 0,
 			'update_site_meta_cache' => false,
 		)
 	);
```
josephfusco, line 226: Nice work leaving `wp_presence_get_network_site_ids()` untouched as activation and deactivation still need to reach archived sites.

josephfusco (approved):
Two nits:

- There is a `get_sites()` in uninstall.php that could benefit from this guard too.

- Switching one test from archived to spam covers all three flags.

presence-api.php
ishitaj34: Thanks! I'll just look into the above suggestions once.

ishitaj34: Hey, I've updated the PR with both suggestions, and updated the description as well to reflect the same.

josephfusco (approved):
🙌🏼

Issue #349, opened 2026-08-24
No end-to-end coverage for any Network Admin presence screen


josephfusco:
`tests/e2e/` holds eight specs and all eight run against a single site. The three Network Admin screens have no browser coverage:

- the Online column on Network Admin > Sites (#335)
- the Online view and column on Network Admin > Users (#336)
- the Who's Online widget on the Network Admin dashboard (#337)

Playwright is not set up to reach a network. `tests/e2e/playwright.config.js` declares one `chromium` project against a single `WP_BASE_URL`, and `.wp-env.json` configures no multisite environment.

All three screens render an avatar stack server-side and then update it over Heartbeat, so keyboard order, focus behaviour across an update, and the accessible name of the overflow control are verified by reading the code and nothing else. `presence-admin-bar-contrast.test.js` is the existing pattern for the contrast half of that.

Fix: a multisite Playwright project and one spec per screen, covering the initial render and one Heartbeat update.

<details>
<summary>Use of AI Tools</summary>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5
Used for: Assisting with review

</details>

i-am-chitti:
@josephfusco Nice to see coverage coming on the multisite side too. I'm picking this up.

josephfusco:
Flagging that this one gates 0.2.0. The milestone is down to #349, #351 and #360, and no multisite surface ships without browser coverage.

The harness is most of the work and none of it exists yet: `.wp-env.json` declares no multisite environment, `tests/e2e/playwright.config.js` runs a single project against one `WP_BASE_URL`, and all eight specs under `tests/e2e/` are single-site. The three specs on top are the easy half.

Worth splitting? One PR for the multisite wp-env environment and a second Playwright project, then one per screen, would get the infrastructure reviewed and landed while the specs are still being written. Say which way you want it, or if you would rather hand off a piece.

josephfusco:
Correcting the note above: this no longer gates 0.2.0. The release is being tagged with the multisite surfaces as they are, and this moves to 0.2.1.

The harness question still stands, and it is still the piece worth settling before any specs get written.

i-am-chitti:
Going with your split: harness first, then one PR per screen. The harness is done and working on its own branch.

One decision to flag before review. I used a second wp-env instance (`tests/e2e/multisite/.wp-env.json`, port 8890) instead of adding `"multisite": true` to the shared `.wp-env.json`:

[...]

PR #397, opened 2026-08-28
test: add a multisite environment for Network Admin e2e coverage

i-am-chitti:
## Description

Part of #349

The environment the Network Admin specs need. The three spec PRs follow, one per screen.

`scripts/start-multisite-env.sh` brings up a second wp-env instance on :8890, network-activates the plugin, creates a `team` sub-site and seeds two users onto both sites. Playwright gets a `chromium-multisite` project matching `presence-network-*.test.js` against that base URL, and `globalSetup` authenticates once per storage state rather than once against `projects[0]`, since the network is a separate install with its own cookies. Why a separate instance rather than a multisite flag on the shared `.wp-env.json` is in the script's header comment.

Verified on a cold start:

| check | result |
| --- | --- |
| tables after `npm run env:start:multisite` | `wp_presence`, `wp_2_presence`, `wp_presence_network_summary` |
| presence seeded on `/team/` | reaches the summary as `[blog_id] => 2`, `[user_count] => 1` |
| a spec on `chromium-multisite` | reaches `/wp-admin/network/` already authenticated |
| single-site project | unchanged, still 30 tests in 8 files |

The E2E job starts the instance and runs no network specs against it yet - those arrive with the next PR - so this run only proves it comes up on a clean runner.

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5
Used for: Investigation, implementation, and tests.

</details>


josephfusco (approved):
Having a separate multisite instance for e2e makes sense, especially with the additional network table.

PR #402, opened 2026-08-28
test: cover the Network Admin Sites list Online column

i-am-chitti:
## Description

Part of #349. Follows #397.

Covers the Online column on Network Admin > Sites: the avatar stack and count for a site with someone online, the em dash for a site with nobody, and the column holding its page-load snapshot across a Heartbeat tick before redrawing on reload. That last one is the deliberate choice documented on `wp_presence_render_network_sites_column()`, and nothing checked it until now.

`clearNetworkPresence()` in the shared helpers re-pushes each site's summary after deleting its rows. Deleting alone leaves a cleared site in the summary until its row ages out, and the screen keeps drawing people who have gone.

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

</details>

### DIFF
--- a/tests/e2e/network-helpers.js
+++ b/tests/e2e/network-helpers.js
@@ -1,0 +1,149 @@
+/**
+ * Presence API — helpers shared by the Network Admin specs.
+ *
+ * The other specs each carry their own copy of `wpCli`. These do not, because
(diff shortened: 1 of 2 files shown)


josephfusco (approved):
Thanks for adding these!

PR #406, opened 2026-08-28
fix: count Who's Online overflow from the heartbeat total

### LINKED ISSUE #360: Who's Online overflow count saturates at 20 and the summary link is unreachable
Who's Online caps its heartbeat payload at 23 entries ([`class-wp-presence-widget-whos-online.php:757`](https://github.com/WordPress/presence-api/blob/1f7bd622d8ec9e31506478a592bf01954079c744/includes/widgets/class-wp-presence-widget-whos-online.php#L757)) but tests `overflow.length > 20` ([`:420`](https://github.com/WordPress/presence-api/blob/1f7bd622d8ec9e31506478a592bf01954079c744/includes/widgets/class-wp-presence-widget-whos-online.php#L420)), so the count saturates at "+20 more" and the summary-link branch never runs.

jigneshbhavani:
The load render is correct and the first tick regresses it. `render()` branches on the full overflow list, so 50 people online loads as `+47 more — view all users`. The tick response caps at `VISIBLE_ROWS + OVERFLOW_THRESHOLD`, the client splits 3 visible rows off that capped array, and `overflow.length > overflowThreshold` can never be true afterwards, so the link is replaced by a `+19 more` expandable list and 28 people become unreachable.

Both the branch and the count now come from `presence-online-total`, which was already in the response and unread.

**The decision this needed that the issue does not cover.** `presence-online-total` counted the viewer, while `render()` and the widget JS both filter the viewer out, so the total was one too high and could not be used as it stood. Rather than subtracting one on the client, `heartbeat_received()` now filters the viewer the same way `render()` does, through a shared `without_current_user()`. Two consequences worth your review:

- `presence-online` no longer carries the viewer's own entry. Three existing tests asserted that it did, including one where the viewer was the only person in the room. They now assert the same payload shape against a second user.
- It also fixes the cap boundary. The cap ran before the viewer was filtered, so at exactly 23 other users the client received 22 and the expanded list silently dropped one.

The client signature now leads with the total. Without that, a user leaving from below the cap moves the count while the capped entries stay identical, and the widget keeps showing the old number.

One correction to the issue body: the count saturates at 19, not 20, for the same off by one reason.

Verified on a local site with 50 users online. Before: server render `+47 more — view all users`, first tick payload `entries: 23, total: 51`, widget then shows `+19 more` with the link gone. After: same server render, payload `total: 50`, widget still shows `+47 more — view all users`.

PHPUnit goes 401 to 403, both green. The new e2e spec covers the tick, and its second phase covers the stale count. Each new guard was mutation checked.

Fixes #360

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5
Used for: Implementation, the PHPUnit and e2e tests, and the local reproduction rig. I reviewed the change and checked the before and after on a local site before opening this.

</details>

### DIFF
--- a/tests/e2e/presence-widgets.test.js
+++ b/tests/e2e/presence-widgets.test.js
@@ -223,4 +223,74 @@ test.describe( 'Presence Widgets', () => {
 		);
 		await expect( postLink ).toBeFocused();
 	} );
+	/**
+	 * The heartbeat payload is capped well below a large room, so the overflow
+	 * count and its link to the Users screen can only come from the total the
+	 * response carries. Reading them off the capped array instead replaces a
+	 * correct server render with a smaller count on the first tick.
+	 */
+	test( "Who's Online counts the overflow from the payload total", async ( {
+		admin,
+		page,
+	} ) => {
+		const others = 26;
+		const visibleRows = 3;
+
+		wpCli(
+			`eval-file wp-content/plugins/presence-api/tests/e2e/whos-online-overflow-seeder.php ${ others }`
+		);
+
+		try {
+			await page.addInitScript( () => {
+				window.__presenceTicks = 0;
+				document.addEventListener( 'DOMContentLoaded', () => {
+					jQuery( document ).on( 'heartbeat-tick', ( event, data ) => {
+						if ( data[ 'presence-online' ] ) {
+							window.__presenceTicks++;
+						}
+					} );
+				} );
+			} );
+
+			await admin.visitAdminPage( '/' );
+
+			const summary = page.locator(
+				'#presence-whos-online-list a.presence-overflow-toggle'
+			);
+
+			const tick = async () => {
+				const before = await page.evaluate( () => window.__presenceTicks );
+				await page.evaluate( () => wp.heartbeat.connectNow() );
+				await page.waitForFunction(
+					( seen ) => window.__presenceTicks > seen,
+					before,
+					{ timeout: 30_000 }
+				);
+			};
+
+			await tick();
+
+			await expect( summary ).toContainText( `+${ others - visibleRows } more` );
+			await expect(
+				page.locator( '#presence-whos-online-list #presence-overflow-list' )
+			).toHaveCount( 0 );
+
+			// Drop a user ranked below the payload cap: the entries the client
+			// receives are byte-identical and only the total moves, so a
+			// signature built from the entries alone leaves the count stale.
+			wpCli(
+				'eval-file wp-content/plugins/presence-api/tests/e2e/whos-online-overflow-seeder.php drop-oldest'
+			);
+
+			await tick();
+
+			await expect( summary ).toContainText(
+				`+${ others - visibleRows - 1 } more`
+			);
+		} finally {
+			wpCli(
+				'eval-file wp-content/plugins/presence-api/tests/e2e/whos-online-overflow-seeder.php clean'
+			);
+		}
+	} );
 } );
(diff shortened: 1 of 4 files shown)


josephfusco (changes requested):
Fix itself is right, but the e2e failure is this spec's own: `tick()` only counts ticks that carry `presence-online`, so it times out whenever the page-load tick wins its leader-election race and the forced tick comes back `presence-online-unchanged` (moving the seeder `wpCli` call inside the `try`, after `visitAdminPage`, makes the next tick a state change either way), and separately `test_heartbeat_payload_holds_every_other_user_at_the_expandable_maximum` still passes with `array_slice( $entries, ... )` reverted because `add_user_to_room( 'dashboard', 0 )` ties every timestamp so the viewer sorts last, which `$i + 1` fixes.

tests/e2e/presence-widgets.test.js
```diff
@@ -223,4 +223,74 @@ test.describe( 'Presence Widgets', () => {
...
+	 */
+	test( "Who's Online counts the overflow from the payload total", async ( {
+		admin,
+		page,
+	} ) => {
+		const others = 26;
+		const visibleRows = 3;
+
+		wpCli(
+			`eval-file wp-content/plugins/presence-api/tests/e2e/whos-online-overflow-seeder.php ${ others }`
```
josephfusco: The room is seeded before the page loads, so the page-load tick can already carry `presence-online` and the forced one comes back `presence-online-unchanged`. `__presenceTicks` never moves and `tick()` waits out its 30 seconds. That is the Playwright failure, not the fix.

tests/widgets/test-widget-whos-online.php
```diff
@@ -260,8 +264,47 @@ public function test_heartbeat_payload_is_capped_to_expandable_list_max() {
...
+	 *
+	 * @covers WP_Presence_Widget_Whos_Online::heartbeat_received
+	 */
+	public function test_heartbeat_payload_holds_every_other_user_at_the_expandable_maximum() {
+		wp_set_current_user( self::$editor_id );
+
+		$others = WP_Presence_Widget_Whos_Online::VISIBLE_ROWS + WP_Presence_Widget_Whos_Online::get_overflow_threshold();
+
+		for ( $i = 0; $i < $others; $i++ ) {
+			$this->add_user_to_room( 'dashboard', 0 );
```
josephfusco: These all share a `date_gmt`, which is what `wp_get_presence()` orders by, so where the viewer lands among them is a tie this is relying on. Revert `$others` to `$entries` in `heartbeat_received()` and it still passes.

tests/e2e/presence-widgets.test.js
jigneshbhavani: @josephfusco Confirmed. Seeding moved inside the `try`, after `visitAdminPage`.

Worth more than a green run, since the old arrangement passed 5 of 5 locally: forcing a real tick to land before the spec's own reproduces CI exactly. The old arrangement times out, the new one passes.

tests/widgets/test-widget-whos-online.php
jigneshbhavani: @josephfusco Confirmed and measured: 24 rows carry 2 distinct `date_gmt` values, and the viewer landed at index 2 here rather than last, which is why the mutation was caught on this machine and not on yours. Either way the guard was resting on the tie.

Now `add_user_to_room( 'dashboard', $i + 1 )`, so the viewer is newest by at least a second and sorts first by construction. Re-ran the mutation: reverting `$others` to `$entries` fails the test again, deterministically this time.

josephfusco (approved):
Thanks for the adjustments. Looks good!

PR #407, opened 2026-08-29
test: cover the Network Admin Users list Online view and column

i-am-chitti:
## Description

Part of #349. Follows #402.

Covers the Online view, its filter and the Online column on Network Admin > Users: the count of everyone online across the network, the column naming a user's site, both sites for a user online on two of them, the view filtering the table down to online users, and the same page-load snapshot behaviour as the Sites column.

The counts include the admin driving the test: rendering any admin screen writes the viewer into the admin room server-side, so the main site always has someone on it. That is why the count reads `(2)` for one seeded user, and the spec says so where it asserts it.

With `wp_presence_get_network_online_user_ids()` stubbed empty and the column's site lookup stubbed out, exactly these five fail and the Sites column tests stay green.

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

</details>

### DIFF
--- a/tests/e2e/network-helpers.js
+++ b/tests/e2e/network-helpers.js
@@ -55,6 +55,19 @@ export function siteUrl( slug = '' ) {
 }
 
 /**
+ * Returns the label the Network Users "Online" column prints for a site.
+ *
+ * The column names sites as domain plus path, which on a subdirectory network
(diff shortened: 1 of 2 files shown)


josephfusco (approved):
🙌🏼

PR #416, opened 2026-08-29
fix: preserve accessible names across heartbeat re-renders in the network Who's Online widget

### LINKED ISSUE #349: No end-to-end coverage for any Network Admin presence screen

i-am-chitti:
## Description

Fixes #349. Follows #407.

`buildListHtml` dropped two things the server render has: the list's `aria-label`, and the wording of the overflow link, which went from `+2 more sites — view all` to `+2 — view all`. The rebuild runs on load through `connectNow()`, not only on some later tick, so this is what a screen reader gets within a second of opening the network dashboard. #390 fixed focus in the same function; this is the rest of it.

Both plural forms now come from the server and the count picks between them, the way the single-site widget already handles `hourAgo` and `hoursAgo` - `_n()` cannot be called from there.

| Before | After |
| --- | --- |
| network-widget-before | network-widget-after |

The link is the visible half. The list's name only shows in the accessibility tree, from the same two runs:

```
- list:                        - list "Sites with online users":
- link "+2 — view all"         - link "+2 more sites — view all"
```

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5
Used for: Investigation, implementation, and tests.

</details>

### DIFF
--- a/includes/widgets/class-wp-presence-network-widget-whos-online.php
+++ b/includes/widgets/class-wp-presence-network-widget-whos-online.php
@@ -231,7 +231,11 @@ class WP_Presence_Network_Widget_Whos_Online {
 		$i18n_json = wp_json_encode(
 			array(
 				'noUsersOnline' => __( 'No users are currently online anywhere on the network.', 'presence-api' ),
-				'viewAll'       => __( 'view all', 'presence-api' ),
+				'sitesOnline'   => __( 'Sites with online users', 'presence-api' ),
+				/* translators: %d: Number of additional sites with online users. */
+				'moreSite'      => __( '+%d more site — view all', 'presence-api' ),
+				/* translators: %d: Number of additional sites with online users. */
+				'moreSites'     => __( '+%d more sites — view all', 'presence-api' ),
 			)
 		);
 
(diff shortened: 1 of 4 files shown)


josephfusco (approved):
Nice work!

PR #424, opened 2026-08-29
fix: count everyone present, including yourself, on every surface

### LINKED ISSUE #423: Presence counts disagree across surfaces
The same question gets two answers depending on the surface. On one page load in Network Admin, the Sites list "Online" column reads `17` for the main site while the admin bar reads `16 online`. Both are correct. The column counts everyone present, the admin bar subtracts the current user.

## Current state

| Surface | Counts you | Where |
| --- | --- | --- |
| Users list count and filter | Yes, appends you if absent | `includes/user-list.php:30-34`, `:83-87` |
| Network Sites "Online" column | Yes | `includes/network-sites-list.php:62-83` |
| Admin bar flyout | Yes, listed first as "(you)" | `includes/admin-bar.php:192-200` |
| Admin bar count and avatar stack | No | `includes/admin-bar.php:26-31`, `:136-152` |
| Who's Online widget, PHP | No | `includes/widgets/class-wp-presence-widget-whos-online.php:626-632` |
| Who's Online widget, JS | No | `includes/widgets/class-wp-presence-widget-whos-online.php:475-476` |

`user-list.php` explicitly appends the current user when missing, so "everyone" is already a deliberate convention here. The split looks like drift.

The admin bar also contradicts itself: the flyout lists you first under "On this page", so the button reads `16` above a list of 17 people.

## Rule

Source returns everyone. Surfaces subtract by identity, never by arithmetic.

`wp_get_presence()` already returns the full set including the current user, and every surface filters at render. Nothing is lost by counting everyone; a surface wanting "others" keeps its `array_filter` on `user_id`. A blind `- 1` is wrong whenever your own row has aged past `WP_PRESENCE_DEFAULT_TTL` on a screen that is not writing heartbeats, because the total is already others-only.

## Tasks

- [ ] Admin bar: count everyone, include your avatar in the stack, update the label and `aria-label` to match
- [ ] Who's Online widget: drop the self filter in both the PHP render and the heartbeat JS, update the empty-state string
- [ ] Keep the visibility gate at `admin-bar.php:33` keyed on others, so a lone user gets no node, and comment why the gate and the count use different sets
- [ ] Remove `$here_count` and `$elsewhere_count` at `includes/admin-bar.php:133-134`, both assigned and never read
- [ ] Update `tests/test-admin-bar.php` and the widget tests, which encode the current contract

Out of scope: `usePresenceUsers` in `src/hooks/use-presence-users.js` keeps `includeSelf = false`. It is a documented public API with an explicit opt-in, and flipping the default would break consumers.

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

Used for: Assisting with mapping the count conventions across surfaces

</details>

josephfusco:
Presence counts disagreed across surfaces because the admin bar and the Who's Online widget filtered out the current user while the users list, the network Sites column and the admin bar flyout did not.

Fixes #423

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

Used for: Assisting with the count convention and its test coverage

</details>

### DIFF
--- a/includes/admin-bar.php
+++ b/includes/admin-bar.php
@@ -137,7 +137,9 @@ function wp_presence_admin_bar_node( $wp_admin_bar ) {
 	foreach ( $here as $entry ) {
 		$stack_ids[] = (int) $entry->user_id;
 	}
-	$stack_ids = array_slice( array_unique( $stack_ids ), 0, 10 );
+	// You, plus nine others. The cap counts you now that you are in the stack.
+	$stack_limit = 10;
+	$stack_ids   = array_slice( array_unique( $stack_ids ), 0, $stack_limit );
 
 	$stack_html = '<span class="presence-bar-avatars">';
 	$z          = count( $stack_ids );
--- a/includes/admin-bar.php
+++ b/includes/admin-bar.php
@@ -130,8 +130,6 @@ function wp_presence_admin_bar_node( $wp_admin_bar ) {
 		}
 	}
 
-	$current_user = get_userdata( $current_uid );
-
 	// You first, then others on this page. Mirrors the flyout's "On this page".
 	$stack_ids = array( $current_uid );
 	foreach ( $here as $entry ) {
@@ -193,6 +191,8 @@ function wp_presence_admin_bar_node( $wp_admin_bar ) {
 	);
 
 	// Current user first.
+	$current_user = get_userdata( $current_uid );
+
 	$wp_admin_bar->add_node(
 		array(
 			'parent' => 'presence-online',
--- a/includes/admin-bar.php
+++ b/includes/admin-bar.php
@@ -153,13 +153,7 @@ function wp_presence_admin_bar_node( $wp_admin_bar ) {
 
 	$stack_html .= '</span>';
 
-	// Subtract by identity, never arithmetic: your row is absent on screens
-	// that are not writing heartbeats, so the total is already others-only.
-	$online_ids = array_map( 'intval', array_unique( wp_list_pluck( $entries, 'user_id' ) ) );
-	if ( $current_uid && ! in_array( $current_uid, $online_ids, true ) ) {
-		$online_ids[] = $current_uid;
-	}
-	$online_count = count( $online_ids );
+	$online_count = count( wp_presence_online_user_ids( $entries ) );
 
 	/* translators: %d: Number of online users, including the current user. */
 	$label = sprintf( _n( '%d online', '%d online', $online_count, 'presence-api' ), $online_count );
--- a/includes/admin-bar.php
+++ b/includes/admin-bar.php
@@ -10,7 +10,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 }
 
 /**
- * Adds a presence indicator to the admin bar showing other online users.
+ * Adds a presence indicator to the admin bar showing online users.
  *
  * @param WP_Admin_Bar $wp_admin_bar The admin bar instance.
  */
@@ -22,7 +22,8 @@ function wp_presence_admin_bar_node( $wp_admin_bar ) {
 	$entries     = wp_get_presence( wp_presence_admin_room() );
 	$current_uid = get_current_user_id();
 
-	// Only show the admin bar node when other users are online.
+	// The count includes you, the gate does not. A lone user gets no node,
+	// but once anyone else is here the number matches every other surface.
 	$others = array_filter(
 		$entries,
 		function ( $e ) use ( $current_uid ) {
@@ -129,19 +130,20 @@ function wp_presence_admin_bar_node( $wp_admin_bar ) {
 		}
 	}
 
-	// "Here" count includes you.
-	$here_count      = count( $here ) + 1; // +1 for current user.
-	$elsewhere_count = count( $elsewhere );
-
-	// Build avatar stack from users on this page only (not current user), capped at 10.
-	$stack_html   = '<span class="presence-bar-avatars">';
 	$current_user = get_userdata( $current_uid );
-	$stack_limit  = 10;
-	$stack_show   = array_slice( $here, 0, $stack_limit );
-	$z            = count( $stack_show );
 
-	foreach ( $stack_show as $entry ) {
-		$user = get_userdata( $entry->user_id );
+	// You first, then others on this page. Mirrors the flyout's "On this page".
+	$stack_ids = array( $current_uid );
+	foreach ( $here as $entry ) {
+		$stack_ids[] = (int) $entry->user_id;
+	}
+	$stack_ids = array_slice( array_unique( $stack_ids ), 0, 10 );
+
+	$stack_html = '<span class="presence-bar-avatars">';
+	$z          = count( $stack_ids );
+
+	foreach ( $stack_ids as $stack_uid ) {
+		$user = get_userdata( $stack_uid );
 		if ( ! $user ) {
 			continue;
 		}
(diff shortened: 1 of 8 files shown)


includes/admin-bar.php
```diff
@@ -151,10 +153,16 @@ function ( $e ) use ( $current_uid ) {
 
 	$stack_html .= '</span>';
 
-	$others_count = count( $others );
+	// Subtract by identity, never arithmetic: your row is absent on screens
+	// that are not writing heartbeats, so the total is already others-only.
+	$online_ids = array_map( 'intval', array_unique( wp_list_pluck( $entries, 'user_id' ) ) );
+	if ( $current_uid && ! in_array( $current_uid, $online_ids, true ) ) {
+		$online_ids[] = $current_uid;
+	}
```
i-am-chitti: Code is repeated - [`user-list.php:27-32`](https://github.com/WordPress/presence-api/blob/27d39c8/includes/user-list.php#L27-L32) and [`:80-85`](https://github.com/WordPress/presence-api/blob/27d39c8/includes/user-list.php#L80-L85) have the others. Better to extract it to a helper like `wp_presence_online_user_ids( $entries )`?

Also `array_map( 'intval', array_unique( ... ) )` dedupes before the cast. It's fine today since `wp_list_pluck` returns strings from the DB, but `array_unique( array_map( 'intval', ... ) )` would be a safer order.

includes/admin-bar.php
```diff
@@ -151,10 +153,16 @@ function ( $e ) use ( $current_uid ) {
 
 	$stack_html .= '</span>';
 
-	$others_count = count( $others );
+	// Subtract by identity, never arithmetic: your row is absent on screens
+	// that are not writing heartbeats, so the total is already others-only.
```
i-am-chitti: I think this comment ended up describing the opposite of the code — the block appends you, but the comment says "subtract" and "the total is already others-only". Looks like it came over from the issue's rule section, where it was explaining why `- 1` is wrong.

Something like this, maybe?

```php
// Your own row can be absent: idle tabs stop pinging and the row ages past
// the TTL. Add yourself by identity when missing, never `+ 1`, which
// double-counts whenever the row is there.
```

tests/test-admin-bar.php
```diff
@@ -222,8 +222,42 @@ public function test_users_on_the_same_admin_page_are_grouped_here() {
...
+		$nodes = $this->render_nodes();
+
+		$this->assertStringContainsString( '3 online', $nodes['presence-online']->title );
+		$this->assertSame( '3 users online', $nodes['presence-online']->meta['aria-label'] );
+	}
+
+	/**
+	 * Your row is absent on screens that never ping, so the total is already
+	 * others-only and must not be decremented again.
+	 */
```
i-am-chitti, line 271: Nothing actually expires in this one. The row is just never written. The docblock says "screens that never ping", which sits a bit oddly against the name.

The TTL case is the one the fix is really for, and I don't think it's covered anywhere yet. `age_entry()` in the widget tests backdates a row past `WP_PRESENCE_DEFAULT_TTL` if you wanted to reuse it.

Also both new tests only assert on the admin bar. Would a test that renders two surfaces and checks they match be worth adding? That's the shape that would have caught the widget gap.

includes/admin-bar.php
```diff
@@ -129,19 +130,20 @@ function ( $e ) use ( $current_uid ) {
 		}
 	}
 
-	// "Here" count includes you.
-	$here_count      = count( $here ) + 1; // +1 for current user.
-	$elsewhere_count = count( $elsewhere );
-
-	// Build avatar stack from users on this page only (not current user), capped at 10.
-	$stack_html   = '<span class="presence-bar-avatars">';
 	$current_user = get_userdata( $current_uid );
```
i-am-chitti: NIt: `$current_user` is accessed [here](https://github.com/WordPress/presence-api/blob/27d39c865942eb7bb2cb5055ea3b3c00d3474ca5/includes/admin-bar.php#L204) but declared much earlier.

We should move it down to near its usage.

readme.txt
```diff
@@ -24,7 +24,7 @@ On a multisite network, Network Admin gets its own view of the same data: a Who'
 
 * Who's Online dashboard widget with idle detection
 * Active Posts dashboard widget grouped by post
-* Admin bar indicator showing other users on the same page
+* Admin bar indicator showing who is on the same page
```
i-am-chitti: The count covers everyone in the room.

```suggestion
* Admin bar indicator showing who's online, grouped by who's on this page
```

includes/widgets/class-wp-presence-widget-whos-online.php
```diff
@@ -622,21 +618,14 @@ private static function render_user_row( $entry, $user ) {
...
-		$current_uid = get_current_user_id();
-		$entries     = array_values(
-			array_filter(
-				$entries,
-				function ( $e ) use ( $current_uid ) {
-					return (int) $e->user_id !== $current_uid;
-				}
-			)
-		);
+		// Lists everyone present, you included, so the count agrees with the
```
i-am-chitti: I don't think the counts line up on first paint. Dropping the self-filter gets us most of the way, but the admin bar and users list also append you when your row is absent, and `render()` doesn't. So the widget can list one fewer person than the count directly above it:

```
your row absent  | bar: 3 online | widget rows: 2 | users list: Online (3)
after heartbeat  | bar: 3 online | widget rows: 3 | users list: Online (3)
```

The heartbeat path is fine. The priority 9 write means your row is always there by the time the payload is built. It's just the server-side render that can be short.

Two ways I see: append here too, which would make the [No users are online](https://github.com/WordPress/presence-api/blob/27d39c8/includes/widgets/class-wp-presence-widget-whos-online.php#L627-L628) unreachable and let it go; or keep it as-is and reword the comment to say the widget lists rows only.

includes/admin-bar.php
i-am-chitti: I think this comment ended up describing the opposite of the code. The block appends the current user, but the comment says "subtract" and "the total is already others-only". Looks like it came over from the issue's rule section, where it was explaining why `- 1` is wrong.

Something like this, maybe?

```php
// Your own row can be absent: idle tabs stop pinging and the row ages past
// the TTL. Add yourself by identity when missing, never `+ 1`, which
// double-counts whenever the row is there.
```

includes/admin-bar.php
```diff
@@ -129,19 +130,20 @@ function ( $e ) use ( $current_uid ) {
...
-	$z            = count( $stack_show );
 
-	foreach ( $stack_show as $entry ) {
-		$user = get_userdata( $entry->user_id );
+	// You first, then others on this page. Mirrors the flyout's "On this page".
+	$stack_ids = array( $current_uid );
+	foreach ( $here as $entry ) {
+		$stack_ids[] = (int) $entry->user_id;
+	}
+	$stack_ids = array_slice( array_unique( $stack_ids ), 0, 10 );
```
i-am-chitti: `$stack_limit = 10` got lost. The cap includes the current user now, so the stack shows 9 others rather than 10 - intended, I think. Reuse `AVATAR_STACK_MAX`, or keep a named local?

i-am-chitti (commented):
One thing worth deciding before this lands: the Who's Online widget drops the self-filter but doesn't add the self-append the admin bar and users list use, so on first paint it can list one fewer person than the count above it. Might be deliberate given the empty-state string in the issue's task list.

josephfusco:
@copilot resolve the merge conflicts in this pull request

josephfusco:
Thanks, all six are addressed as of ad60bd7, including option A on the widget and the cross-surface test.

josephfusco:
Merged this one out of habit before you had a second look, so flag anything you want changed and it goes in a follow-up.

PR #426, opened 2026-08-30
perf: store network summary rows compact

### LINKED ISSUE #329: Network summary rows are stored pretty-printed, at 2x the bytes

josephfusco:
The `data` column is selected in full for every site on each network-admin request, and the measurement on #329 put pretty-printing at twice the stored and transferred bytes for no latency difference.

Fixes #329

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

Used for: Assisting with the change and running the multisite suite

</details>

### DIFF
--- a/includes/network-functions.php
+++ b/includes/network-functions.php
@@ -310,11 +310,10 @@ function wp_presence_record_network_summary_push( array $user_ids ) {
  * Shaped to be read in a database client, the only place this column is looked
  * at directly:
  *
- *     {
- *         "site": "example.com/shop/",
- *         "scheme": "https",
- *         "online_user_ids": [ 3, 7 ]
- *     }
+ *     {"site":"example.com/shop/","scheme":"https","online_user_ids":[3,7]}
+ *
+ * Three keys stay legible compact, and the read selects this column in full for
+ * every site on each network-admin request, so pretty-printing doubled it.
  *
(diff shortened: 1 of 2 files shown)


i-am-chitti (approved):
LGTM 🚀

Issue #400, opened 2026-08-28
No site-level control over whether presence is recorded


josephfusco:
Presence is recorded for every logged-in user on every admin screen, and the site owner has no supported way to turn it off. This issue was filed as a per-user opt-out; that is the wrong control, for the reasons below, and the scope here is now a controller-level switch.

## What exists today

| Layer | Surfaces | Control |
| --- | --- | --- |
| Presence recording | Admin bar, Who's Online, Active Posts, post lock bridge, REST, CLI | None. The write at `includes/functions.php:108` is ungated |
| Network aggregate | Sites list Online column, Users list Online view and column, network dashboard widget | `wp_presence_network_aggregation_enabled` (`includes/network-functions.php:86-102`) |

That filter is not the privacy lever it looks like. It is `@access private`, it defaults to `! wp_is_large_network()`, and a default that turns on the size of the network is answering a write-concentration question, not a policy one.

## Acceptance criteria

- [ ] A site owner can stop presence rows being written, and every surface agrees while it is off.
- [ ] The network aggregate can be switched off independently of recording, through public API rather than the private performance filter.
- [ ] Whatever the control is, filter or option or both, it carries a `@since` and is named in the privacy policy content from #393.
- [ ] The default is stated, and defensible against [Art. 25(2)](https://gdpr-info.eu/art-25-gdpr/).
- [ ] Precedence between a network admin's choice and a site admin's is defined on a network-activated install.

Admin UI and WP-CLI surfaces for the switch are tracked in #441.

## Open questions

- Filter only, or a stored setting, and if a setting, which screen it lives on.
- Whether recording and aggregation are one control with two levels or two independent switches.

## Worth knowing

Rows expire after 150 seconds (`WP_PRESENCE_DEFAULT_TTL`), so stopping the write clears every screen within one TTL. There is no separate "hide me from other people's screens" behavior to design.

<details>
<summary><b>Why not a per-user opt-out</b></summary>

Presence is a negative signal: the value is that nobody else is on the post. If a user can hide, absence stops meaning anything, and the post lock bridge is where that turns into two people overwriting each other. Core takes the same position already: `wp_check_post_lock()` broadcasts a display name and avatar to anyone with edit rights and has no opt-out.

[Art. 21](https://gdpr-info.eu/art-21-gdpr/) is a balancing test against legitimate-interest processing, not an absolute switch, and preventing concurrent overwrites is a compelling ground.

[...]

PR #429, opened 2026-08-30
feat: add a site and network switch for whether presence is recorded

josephfusco:
Settles the three open questions on the issue as a filter rather than a stored setting, recording and aggregation as separate switches, and strictest-wins precedence on multisite.

Refs #400

<details>
<summary>AI disclosure</summary>

Used for: Assisting with the implementation and tests.

</details>

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -2,7 +2,7 @@
 /**
  * Presence API functions.
  *
- * Public API (7 functions):
+ * Public API:
  *   wp_get_presence()
  *   wp_set_presence()
  *   wp_remove_presence()
@@ -10,6 +10,7 @@
  *   wp_can_access_presence_room()
  *   wp_presence_post_room()
  *   wp_presence_admin_room()
+ *   wp_presence_recording_enabled()
  *
  * @package Presence_API
  */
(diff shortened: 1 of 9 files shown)


includes/functions.php
```diff
@@ -95,6 +143,10 @@ function wp_get_presence( $room, $timeout = WP_PRESENCE_DEFAULT_TTL ) {
 function wp_set_presence( $room, $client_id, $state, $user_id = 0 ) {
 	global $wpdb;
 
+	if ( ! wp_presence_recording_enabled() ) {
+		return false;
+	}
```
i-am-chitti, line 148: Returning `false` here makes the REST API create path resolve in 500 HTTP code : ([`controller:497-503`](https://github.com/WordPress/presence-api/blob/485bf3d/includes/class-wp-rest-presence-controller.php#L497-L503)). 

A client can't tell "disabled" from "broke", and 500 invites a retry. Should we have WP_Error return (403 HTTP status) at this layer (no change at API) or a different return value to be identified if its disabled at the API layer? 

Same at [`cli-command.php:70`](https://github.com/WordPress/presence-api/blob/485bf3d/includes/cli/class-wp-presence-cli-command.php#L70).

includes/network-functions.php
```diff
@@ -72,15 +72,23 @@ function wp_presence_has_network_summary_table() {
...
  * rather than having a network inherit that cost from installing the plugin.
+ * That default answers write concentration rather than policy, so a network
+ * with a view either way should set it deliberately.
  *
  * The read path checks this too, or a network that stopped aggregating would go
  * on drawing the rows it already had with nothing to say the feed had stopped.
  *
- * @access private
+ * @since 0.2.0
+ * @since 0.3.0 Promoted from an @access private helper to the public API.
```
i-am-chitti, line 90: As being promoted to public API, there's no list it can go in. [the README block](https://github.com/WordPress/presence-api/blob/485bf3dc831aab768b49bf58ce9912529d8411f7/README.md?plain=1#L44-L70) both scope themselves to `includes/functions.php` and not `includes/network-functions.php` . Should we widen the readme?

tests/test-functions.php
```diff
@@ -823,4 +823,51 @@ public function test_avatar_stack_skips_a_user_with_no_avatar() {
...
 	}
+
+	/**
+	 * Asserted at the table rather than only on the return value: a switch that
+	 * reports false while the row still lands is the failure that matters.
+	 *
+	 * @covers ::wp_presence_recording_enabled
+	 * @covers ::wp_set_presence
+	 */
+	public function test_switching_recording_off_writes_nothing() {
```
i-am-chitti, line 843: default test case is missing though issue's AC include - 

> The default is stated, and defensible against

includes/functions.php
josephfusco: Right, and the CLI had the same conflation. Both now check `wp_presence_recording_enabled()` first and report `rest_presence_recording_disabled` with 501, matching core's [`application_passwords_disabled`](https://github.com/WordPress/wordpress-develop/blob/a5f6b9c393071f3f3a2b24d161e57c391ec5d1b4/src/wp-includes/rest-api/endpoints/class-wp-rest-application-passwords-controller.php#L686), which also keeps it distinct from the 503 for a missing table.

includes/network-functions.php
josephfusco: Widened it: the `@access private` sentence now names `includes/network-functions.php` too, and there is a `### Network` subsection under `## PHP API` for the function, mirroring the one under `## REST API`.

tests/test-functions.php
josephfusco: Added, `test_recording_is_on_with_no_filters_added`.

Issue #403, opened 2026-08-28
e2e specs hit wordpress.org on every WP-CLI call, so one timeout fails the job


i-am-chitti:
From https://github.com/WordPress/presence-api/pull/374#pullrequestreview-5040138151, where `e2e / Playwright` went red on a PR that only touched `.distignore`.

Four specs run WP-CLI by shelling out to `npx wp-env run cli wp ...`, and they do it dozens of times per run:

- [`presence-screenshots.test.js#L21`](https://github.com/WordPress/presence-api/blob/34b520c4d530e339a6fb1f93cf58831007883e14/tests/e2e/presence-screenshots.test.js#L21)
- [`presence-widgets.test.js#L19`](https://github.com/WordPress/presence-api/blob/34b520c4d530e339a6fb1f93cf58831007883e14/tests/e2e/presence-widgets.test.js#L19)
- [`presence-admin-bar-contrast.test.js#L17`](https://github.com/WordPress/presence-api/blob/34b520c4d530e339a6fb1f93cf58831007883e14/tests/e2e/presence-admin-bar-contrast.test.js#L17)
- [`presence-stale-screen-coalescing.test.js#L61`](https://github.com/WordPress/presence-api/blob/34b520c4d530e339a6fb1f93cf58831007883e14/tests/e2e/presence-stale-screen-coalescing.test.js#L61)

[`.wp-env.json`](https://github.com/WordPress/presence-api/blob/34b520c4d530e339a6fb1f93cf58831007883e14/.wp-env.json#L2) sets `"core": null`, which means "use the latest WordPress". To find out what the latest is, wp-env asks wordpress.org — once for every call, because each call is a new process and nothing is cached between them:

    $ node --require netspy.js node_modules/@wordpress/env/bin/wp-env run cli wp option get home
    [netspy] dns.resolve   -> WordPress.org
    [netspy] https.request -> https://api.wordpress.org/core/stable-check/1.0/

So a slow moment on wordpress.org fails a test that has nothing to do with the network. And [`playwright.config.js`](https://github.com/WordPress/presence-api/blob/34b520c4d530e339a6fb1f93cf58831007883e14/tests/e2e/playwright.config.js#L44) sets no `retries`, so the default of `0` means one bad call takes the whole job down. That is what happened on [run 33050559661](https://github.com/WordPress/presence-api/actions/runs/33050559661/job/98444763297) — all three failures were this timeout.

### Fix

[...]

PR #435, opened 2026-08-30
fix: add data-post-id to server-rendered Active Posts rows

### LINKED ISSUE #372: Active Posts focus-retention e2e test is flaky
[`tests/e2e/presence-widgets.test.js:188`](https://github.com/WordPress/presence-api/blob/005b6cad3d82c3fcde1867893cdb781f321ce092/tests/e2e/presence-widgets.test.js#L188) is flaky, failing 2 of 9 local runs on `005b6ca` with `expect(locator).toBeFocused()` reporting the post link as `inactive` rather than unfocused, so the element is gone from the re-rendered list instead of having lost focus.

i-am-chitti:
Fixes #372

The Active Posts widget renders its list twice — once server-side in `WP_Presence_Widget_Active_Posts::render()`, then again in `assets/js/active-posts.js` on each heartbeat tick whose signature changed. Only the JS build emitted `data-post-id`, and that attribute is the hook focus retention keys on.

So on the first tick after page load, `captureFocus()` found the focused link inside the container but `.closest('[data-post-id]')` matched nothing, and `restoreFocus()` fell through to `container.trigger('focus')` — moving focus to the wrapper `<div tabindex="-1">` instead of back to the post link. Focus then stayed stuck there, since `$.contains()` is false for the container itself.

Adding the attribute to the server-rendered `<li>` brings it in line with the JS build and with Who's Online, which already emits `data-user-id` on both paths — which is why only Active Posts was affected.

Note - #403: this is a real focus bug, not test noise, so retries would have hidden it rather than fixed it.

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5
Used for: Diagnosing the root cause and verifying the fix against the e2e test

</details>

### DIFF
--- a/includes/widgets/class-wp-presence-widget-active-posts.php
+++ b/includes/widgets/class-wp-presence-widget-active-posts.php
@@ -145,7 +145,7 @@ class WP_Presence_Widget_Active_Posts {
 				// Active is the default state; labeling it adds noise.
 				$status_label = $any_active ? '' : __( 'Idle', 'presence-api' );
 
-				echo '<li class="presence-active-post-item">';
+				echo '<li class="presence-active-post-item" data-post-id="' . (int) $post_data['post_id'] . '">';
 
 				// Avatar stack.
 				echo '<span class="presence-editor-stack">';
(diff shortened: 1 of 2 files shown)


josephfusco (approved):
Nice solution.  This works!

PR #436, opened 2026-08-30
test: run Jest unit tests in CI

### LINKED ISSUE #405: test:unit fails locally and doesn't run in CI
`test:unit` runs Jest, but the repo has no Jest config. So Jest picks up every `*.test.js` file it can find, including the Playwright specs and the `node:test` files under `.github/scripts`. It cannot run either kind. On 34b520c:

    Test Suites: 12 failed, 2 passed, 14 total
    Tests:       17 passed, 17 total

The 2 that pass are the real unit tests. The 12 failures are files Jest was never meant to touch, and every new e2e spec adds one more.

Nothing in [`ci.yml`](https://github.com/WordPress/presence-api/blob/34b520c4d530e339a6fb1f93cf58831007883e14/.github/workflows/ci.yml) runs [`test:unit`](https://github.com/WordPress/presence-api/blob/34b520c4d530e339a6fb1f93cf58831007883e14/package.json#L13), which is why this went unnoticed. It also means those 17 assertions are never checked on a PR.

### Fix

Give the repo a Jest config that limits Jest to the real unit tests, and add a `ci.yml` job that runs `npm run test:unit` so it stays working. Those live under both `src/` and `assets/js/`: #401 adds `assets/js/test/tab-coordinator.test.js`, so scoping the config to `src/` alone would silently drop it.

One thing to watch when implementing: `wp-scripts` only supplies its own Jest preset while the project has no config of its own, so a new config has to carry the preset and Babel transform forward rather than start fresh.

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5
Used for: Assisting with research and drafting

</details>

aldorizona10-glitch:
## Summary

- scope Jest discovery to the unit-test directories under `src/` and `assets/js/`
- retain the `@wordpress/scripts` preset and Babel transform
- run `npm run test:unit` in the pull-request CI workflow

Fixes #405

## Testing

- `git diff --check`
- Local Jest execution could not complete because the local npm installation was incomplete; CI installs dependencies from scratch on Node 22.

### DIFF
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -1,0 +1,22 @@
+name: Unit Tests
+
+on:
+  workflow_call:
+
+permissions:
+  contents: read
+
+jobs:
+  unit:
+    name: Jest
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version: 22
+          cache: npm
+      - run: npm ci
+      - run: npm run test:unit
--- a/jest-unit.config.js
+++ b/jest-unit.config.js
@@ -1,0 +1,11 @@
+const defaultConfig = require( '@wordpress/scripts/config/jest-unit.config' );
+
+module.exports = {
+	...defaultConfig,
+	testMatch: [
+		'<rootDir>/src/**/test/**/*.[jt]s?(x)',
+		'<rootDir>/src/**/*.test.[jt]s?(x)',
+		'<rootDir>/assets/js/test/**/*.[jt]s?(x)',
+		'<rootDir>/assets/js/**/*.test.[jt]s?(x)',
+	],
+};
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,10 @@ jobs:
     needs: gate
     uses: ./.github/workflows/e2e.yml
 
+  unit:
+    needs: gate
+    uses: ./.github/workflows/unit.yml
+
   plugin-check:
     needs: gate
     uses: ./.github/workflows/plugin-check.yml


jest-unit.config.js
```diff
@@ -0,0 +1,11 @@
+const defaultConfig = require( '@wordpress/scripts/config/jest-unit.config' );
+
+module.exports = {
+	...defaultConfig,
+	testMatch: [
+		'<rootDir>/src/**/test/**/*.[jt]s?(x)',
+		'<rootDir>/src/**/*.test.[jt]s?(x)',
+		'<rootDir>/assets/js/test/**/*.[jt]s?(x)',
+		'<rootDir>/assets/js/**/*.test.[jt]s?(x)',
```
i-am-chitti, line 9:
```js
'<rootDir>/src/**/test/**/*.[jt]s?(x)',      // line 6 
'<rootDir>/assets/js/test/**/*.[jt]s?(x)',   // line 8
```

These two globs match every file in a `test/` folder, not just tests. If someone adds a shared helper here later, Jest treats it as a suite and fails with `Your test suite must contain at least one test`.

They're also redundant: all three current test files are already matched by the `*.test.[jt]s?(x)` patterns on lines 7 and 9. Removing lines 6 and 8 keeps the same 3 suites / 21 tests passing. So, we can remove these ones.

.github/workflows/unit.yml
```diff
@@ -0,0 +1,22 @@
...
+on:
+  workflow_call:
+
+permissions:
+  contents: read
+
+jobs:
+  unit:
+    name: Jest
+    runs-on: ubuntu-24.04
```
i-am-chitti, line 12: Let's add timeout too as other CIs -

```suggestion
    runs-on: ubuntu-24.04
    timeout-minutes: 5
```

i-am-chitti (commented):
Two more files outside this diff need updating:

**`.distignore`**
- `jest-unit.config.js` should be added next to [`package.json`](https://github.com/WordPress/presence-api/blob/0f09f746550e6f21846b5ffb968ad1e0cca3b369/.distignore#L25).
- `assets/js/test/` is also absent, while [src/hooks/test/ , src/utils/test/](https://github.com/WordPress/presence-api/blob/0f09f746550e6f21846b5ffb968ad1e0cca3b369/.distignore#L32-L33) are listed. This PR makes `assets/js/test/` a first-class unit-test location, so would be good to add there as well.

**`.github/CONTRIBUTING.md`**
- [Line 78](https://github.com/WordPress/presence-api/blob/0f09f746550e6f21846b5ffb968ad1e0cca3b369/.github/CONTRIBUTING.md#L78) lists the gating checks as "PHPCS, PHPStan, PHPUnit across PHP 7.4 + 8.3 plus a multisite run, Playwright" - Jest should now get an entry in that list.

cc: @josephfusco

i-am-chitti:
@aldorizona10-glitch Thanks for the contribution!

One process note for next time: please comment on the issue before starting work, so a maintainer can assign it to you and we avoid two people building the same thing. [`CONTRIBUTING.md`](https://github.com/WordPress/presence-api/blob/main/.github/CONTRIBUTING.md?plain=1#L34-L42) covers the flow.

For now, could you drop a comment on #405? It's still unassigned, and that lets us assign it to you so the PR is properly linked to the asignee's issue.

josephfusco (commented):
Nice first pass at this, it lands #405 cleanly and the new job comes in green at 35s.

Agreed with all four of Deepak's points, though worth knowing that `assets/js/test/tab-coordinator.test.js` already ships in the built zip today, so that `.distignore` line is closing a leak you did not introduce.

PR #440, opened 2026-08-31
fix: public surface read constants

### LINKED ISSUE #12: Alert when a screen is stale due to a user having updated settings

### LINKED ISSUE #410: Public surface detection misses constants the plugin only reads
`detect-public-surface.js` matches constants only in the guarded define form, so a constant a site sets and the plugin merely reads is never flagged.

```js
// .github/scripts/detect-public-surface.js:23
[/if\s*\(\s*!\s*defined\s*\(\s*(['"])(WP_PRESENCE_[A-Z0-9_]*)\1\s*\)\s*\)/g, (m) => `constant:${m[2]}`],
```

## Evidence

| Constant | Form in the source | Detected |
| --- | --- | :-: |
| `WP_PRESENCE_DEFAULT_TTL` | `if ( ! defined( ... ) ) { define( ... ); }` | Yes |
| `DISABLE_WP_PRESENCE` | `defined( ... ) && DISABLE_WP_PRESENCE` | No |

Both are the same promise. A site puts either one in `wp-config.php` and depends on the name from then on, so renaming either one breaks that site, which is what the script exists to catch.

```console
$ node .github/scripts/detect-public-surface.js main feat/400-site-level-presence-control
Found 1 new public surface:
- `wp_presence_enabled` (filter)
```

That branch adds `DISABLE_WP_PRESENCE` as well.

The prefix in the pattern is also load-bearing today and would not be after a fix: `DISABLE_WP_PRESENCE` leads with the verb, the way `DISABLE_WP_CRON` does, so a `WP_PRESENCE_` prefix match cannot find it.

## Tasks

- [ ] Match the read form for constants the plugin never defines
- [ ] Keep core constants out of the result, `ABSPATH` and `WP_DEBUG` above all, since every file opens with one
- [ ] Cover both forms in `.github/scripts/detect-public-surface.test.js`

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

Used for: Assisting with the audit of the detection rules

</details>

iqbal-web:
Fixes #410 

<details>

AI assistance: Yes
Tool(s): GitHub Copilot, ChatGPT
Model(s): GPT-5.1
Used for: Initial code skeleton and test suggestions; final implementation and tests were reviewed and edited by me.

</details>

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -83,7 +83,7 @@ A new label joins an existing color. If it genuinely needs its own, check it in 
 
 ### Public surfaces
 
-A hook, REST route, WP-CLI command, guarded constant, or browser global that a site can reach is a promise: renaming it later breaks that site. `public-surface.yml` labels any pull request that adds one and comments with the ones nothing has been written about yet.
+A hook, REST route, WP-CLI command, constant, or browser global that a site can reach is a promise: renaming it later breaks that site. `public-surface.yml` labels any pull request that adds one and comments with the ones nothing has been written about yet.
 
 Written about means a docblock directly above it with a sentence saying what it is for. That is where WordPress core's code reference reads from, and it keeps the write-up next to the code instead of in a hand-kept list that drifts the first time someone forgets it. `README.md` narrates the surfaces worth a worked example; it is not the inventory.
 
--- a/.github/scripts/detect-public-surface.js
+++ b/.github/scripts/detect-public-surface.js
@@ -1,6 +1,6 @@
 'use strict';
 
-// A hook, route, guarded constant, CLI command, or browser global is a promise:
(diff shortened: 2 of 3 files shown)


josephfusco (approved):
Nice work with this!

PR #443, opened 2026-08-31
docs: correct overflow threshold docblock

### LINKED ISSUE #437: Who's Online overflow threshold is documented as filterable but no filter exists
`WP_Presence_Widget_Whos_Online::get_overflow_threshold()` says "Returns the overflow threshold, filtered for customization" and returns the constant unchanged. There is no `apply_filters` for it anywhere in the plugin.

`includes/widgets/class-wp-presence-widget-whos-online.php:48-60`

```php
/**
 * Returns the overflow threshold, filtered for customization.
 * ...
 */
public static function get_overflow_threshold() {
	return self::OVERFLOW_THRESHOLD;
}
```

The method looks deliberate rather than accidental. `OVERFLOW_THRESHOLD` is the only one of the widget's four constants with a getter at all. `VISIBLE_ROWS`, `IDLE_THRESHOLD` and `AVATAR_STACK_MAX` are each referenced directly. The getter's only reason to exist is the filter the docblock describes.

Nothing was removed. The method and its docblock arrived together in `942be6e`.

The shape the plugin already uses for this is `wp_presence_get_heartbeat_idle_ticks()` at `includes/heartbeat.php:40`, a getter whose whole body is `apply_filters( 'wp_presence_heartbeat_idle_ticks', 5 )` with the filter carrying its own docblock.

Two ways to close it, and which one you want is a call I would rather not guess:

1. Add `wp_presence_whos_online_overflow_threshold` in that same shape. That adds public API surface, so it picks up the `Public API` label and is a commitment to keep.
2. Correct the docblock and leave the threshold fixed.

Happy to do either.

muneeb-ashraf:
Correct the overflow threshold getter docblock to describe the threshold as fixed, matching the current implementation and the maintainer decision.

Fixes #437

### Testing

- `php -l includes/widgets/class-wp-presence-widget-whos-online.php`
- `php vendor/squizlabs/php_codesniffer/bin/phpcs --standard=phpcs.xml.dist includes/widgets/class-wp-presence-widget-whos-online.php`
- `git diff --check`

I used the AI tool "Codex" to understand the issue and get help with the code.

### DIFF
--- a/includes/widgets/class-wp-presence-widget-whos-online.php
+++ b/includes/widgets/class-wp-presence-widget-whos-online.php
@@ -47,7 +47,7 @@ class WP_Presence_Widget_Whos_Online {
 	const AVATAR_STACK_MAX = 4;
 
 	/**
-	 * Returns the overflow threshold, filtered for customization.
+	 * Returns the fixed overflow threshold.
 	 *
 	 * When the number of overflow users exceeds this value, the widget
 	 * switches from an expandable list to a compact summary linking to


josephfusco:
Welcome, and thanks for picking this up. The change reads right, option 2 from the issue, threshold stays fixed.

You did more work than this one needed, though. A Testing section earns its keep when there's behavior to verify, and for a docblock correction lint and CI already cover it. No need to run that checklist by hand on changes like this.

Could you add the template's Use of AI Tools section back to the description? The template asks for it on every PR, and "No AI assistance" is a complete answer if that's the case.

josephfusco:
Merged. Nice first contribution @muneeb-ashraf

Issue #415, opened 2026-08-29
Collaboration hooks are undocumented


josephfusco:
`wp_presence_collaboration_started` and `wp_presence_collaboration_ended` have shipped since 0.1.21 and are absent from the README. The Actions section lists only `wp_presence_screen_revision_bumped`, so a third party reading the documented surface has no way to know these exist.

```php
// includes/heartbeat.php:398, :408
do_action( 'wp_presence_collaboration_started', $room, $entries );
do_action( 'wp_presence_collaboration_ended', $room, $entries );
```

Both pass `$room` and the current `$entries`, and fire on the 1 to 2+ editor edge and back.

WordPress/sync-storage already consumes both, in [`lib/rtc/server-authority.php`](https://github.com/WordPress/sync-storage/blob/d9201aa4972abfdeba22a54c88584f20d6165d8a/lib/rtc/server-authority.php#L20), to announce a room active or inactive without touching post meta. So there is a live external dependency on an undocumented pair.

Related: #417 for the other places that integration works around this plugin's contract.

## Tasks

- [ ] Add both to the Actions section of `README.md`, with the edge they fire on and the transient that debounces them

<details>
<summary><b>Use of AI Tools</b></summary>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

Used for: Assisting with locating the undocumented hooks

</details>

josephfusco:
Closed by #445

PR #445, opened 2026-09-01
docs: Document wp_presence_collaboration_started and wp_presence_collaboration_ended hooks

theaminulai:
### **Why**
The `wp_presence_collaboration_started` and `wp_presence_collaboration_ended` hooks have been shipped since version 0.1.21 but were absent from the README.md documentation. This created a discoverability problem for third-party developers who had no way to know these hooks exist without reading the source code. The WordPress sync-storage plugin already depends on these hooks, making them a critical part of the public API contract. #415

### **What**
Added documentation for two previously undocumented action hooks to the Actions section of README.md:

1. **`wp_presence_collaboration_started`** – Fires when collaboration begins in a room (transition from 1 to 2+ editors)
2. **`wp_presence_collaboration_ended`** – Fires when collaboration ends in a room (transition from 2+ to 1 editor)

Both entries include:
- Clear description of when each hook fires
- Parameter documentation (`$room`, `$entries`)
- Example usage code
- Explanation of transient debouncing mechanism with TTL

### **How**
- Inserted two new hook documentation sections in the **Actions** subsection of README.md, immediately following the existing `wp_presence_screen_revision_bumped` hook documentation
- Maintained consistent formatting and structure with existing documentation
- Included technical details about debouncing via transients to prevent duplicate fires during connection/disconnection churn

### **Test Instructions**

1. **Verify documentation exists**: Open README.md and confirm both `wp_presence_collaboration_started` and `wp_presence_collaboration_ended` appear in the Actions section
2. **Verify formatting**: Ensure the new sections match the existing hook documentation style with:
   - Clear hook name in backticks
   - Description of when it fires
   - PHP code example with `add_action()`
   - Correct parameter count (10, 2)
3. **Verify accuracy**: Cross-reference the documentation against `includes/heartbeat.php` lines 398 and 408 to ensure parameter descriptions match the actual implementation
4. **Verify completeness**: Confirm all three action hooks (including `wp_presence_screen_revision_bumped`) are now documented in a single Actions section

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,7 +157,7 @@ add_action( 'wp_presence_collaboration_started', function( $room, $entries ) {
 ```
 
 #### `wp_presence_collaboration_ended`
-Fires when collaboration ends in a room (transition from 2+ to 1 editor). Debounced via transient with TTL equal to the presence TTL, preventing duplicate fires during disconnection churn.
+Fires when collaboration ends in a room (transition from 2+ to exactly 1 editor). The check runs on an editor heartbeat tick, so if every editor leaves at once there is nobody left to tick and the hook does not fire; the transient expires on the presence TTL and the room resets quietly.
 ```php
 add_action( 'wp_presence_collaboration_ended', function( $room, $entries ) {
     // Announce room inactive or update integration state
--- a/README.md
+++ b/README.md
@@ -148,6 +148,22 @@ add_action( 'wp_presence_screen_revision_bumped', function( $screen_key, $revisi
 }, 10, 3 );
 ```
 
+#### `wp_presence_collaboration_started`
+Fires when collaboration starts in a room (transition from 1 to 2+ editors). Debounced via transient with TTL equal to the presence TTL, preventing duplicate fires during connection churn.
+```php
+add_action( 'wp_presence_collaboration_started', function( $room, $entries ) {
+    // Announce room active or update integration state
+}, 10, 2 );
+```
+
+#### `wp_presence_collaboration_ended`
+Fires when collaboration ends in a room (transition from 2+ to 1 editor). Debounced via transient with TTL equal to the presence TTL, preventing duplicate fires during disconnection churn.
+```php
+add_action( 'wp_presence_collaboration_ended', function( $room, $entries ) {
+    // Announce room inactive or update integration state
+}, 10, 2 );
+```
+
 ## REST API
 
 All endpoints require `edit_posts`. Responses include `Cache-Control: no-store`.
--- a/README.md
+++ b/README.md
@@ -149,7 +149,7 @@ add_action( 'wp_presence_screen_revision_bumped', function( $screen_key, $revisi
 ```
 
 #### `wp_presence_collaboration_started`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -148,6 +148,22 @@ add_action( 'wp_presence_screen_revision_bumped', function( $screen_key, $revisi
 }, 10, 3 );
 ```
 
+#### `wp_presence_collaboration_started`
+Fires when collaboration starts in a room (transition from 1 to 2+ editors). Debounced via transient with TTL equal to the presence TTL, preventing duplicate fires during connection churn.
```
josephfusco: The transient is not a debounce, it is where the previous editor count lives. Every tick is a separate request, so `:379` reads the last count back to detect the edge, and `:409` rewrites it on every tick rather than only on a transition. The comment above it says why: a steady pair that let the transient lapse would read back as a fresh start and re-announce itself.

The TTL half is right. Worth adding that only rows whose `client_id` starts with `editor-` count toward the transition (`:371`), while `$entries` is every entry in the room, so a listener calling `count( $entries )` gets a different number than the one that fired the hook.

```suggestion
Fires when collaboration starts in a room (transition from 1 to 2+ editors). Only entries whose `client_id` begins with `editor-` count toward the transition, while `$entries` is every entry in the room. The previous count is held in a transient that expires on the presence TTL, so once those entries have aged out the next pair reads as a fresh start.
```

README.md
```diff
@@ -148,6 +148,22 @@ add_action( 'wp_presence_screen_revision_bumped', function( $screen_key, $revisi
...
+```php
+add_action( 'wp_presence_collaboration_started', function( $room, $entries ) {
+    // Announce room active or update integration state
+}, 10, 2 );
+```
+
+#### `wp_presence_collaboration_ended`
+Fires when collaboration ends in a room (transition from 2+ to 1 editor). Debounced via transient with TTL equal to the presence TTL, preventing duplicate fires during disconnection churn.
```
josephfusco: Same on the debounce wording, plus one thing worth knowing here. `wp_presence_check_collaboration_threshold()` is called from a single place, `includes/heartbeat.php:334`, inside the editor heartbeat and right after the ticking user's own row is written. There is always at least one editor at check time, so `2+ to 1` is the only exit.

Which means if both editors close the editor at once, nothing ticks that room again, the hook never fires, and the transient just expires. That is the case the example below cares about most, since `// Announce room inactive` will not run when the room actually empties.

```suggestion
Fires when collaboration ends in a room (transition from 2+ to exactly 1 editor). The check runs on an editor heartbeat tick, so if every editor leaves at once there is nobody left to tick and the hook does not fire; the transient expires on the presence TTL and the room resets quietly.
```

josephfusco (commented):
Useful hooks to have written down, and the formatting follows the `wp_presence_screen_revision_bumped` entry above it. Two corrections inline, both from reading `includes/heartbeat.php:334-413`.

theaminulai:
@josephfusco Thanks for pointing these out. I’ve fixed both corrections based on `includes/heartbeat.php:334-413` and updated the relevant entries accordingly.

josephfusco (approved):

Issue #414, opened 2026-08-29
Post-lock bridge's stated intent predates Gutenberg sync awareness


josephfusco:
`post-lock-bridge.php` says the block editor will consume presence data directly in the future. That is settled now, and not the way the header describes it.

```php
// includes/post-lock-bridge.php:7
// The intent is for the block editor (Gutenberg) to consume presence
// data directly in the future, enabling real-time awareness (cursors,
// selections, who's editing which block) rather than the current blunt
// lock/takeover model.
```

The editor consumes `WP_Sync_Storage` behind `__unstable_wp_sync_storage` (WordPress/gutenberg#81697). [WordPress/sync-storage](https://github.com/WordPress/sync-storage/blob/d9201aa4972abfdeba22a54c88584f20d6165d8a/lib/rtc/class-sync-storage-provider.php) implements that interface, delegates `get_awareness_state()` and `set_awareness_state()` to `wp_get_presence()` and `wp_set_presence()`, and keeps CRDT updates in its own `wp_collaboration` table.

The delegation needs no room mapping because both sides already build the same string:

```php
// presence: includes/functions.php:301
'postType/' . $post->post_type . ':' . $post->ID;
```
```js
// gutenberg: packages/sync/src/providers/http-polling/http-polling-provider.ts:141
const room = objectId ? `${ objectType }:${ objectId }` : objectType;
```

So the split is awareness and cursors in the editor to sync-storage, room membership across the rest of wp-admin to this plugin. Nothing in this repo says so.

Related: #417 for what a consumer has to translate to use the delegation, #415 for the hooks it also relies on.

## Tasks

- [ ] Write the split into `README.md`
- [ ] Replace the `post-lock-bridge.php` header with a pointer to it

<details>
<summary><b>Use of AI Tools</b></summary>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

Used for: Assisting with tracing the awareness path across the two repositories

</details>

jigneshbhavani:
@josephfusco #447 landed both boxes here on 2026-09-02 (`b045eb0`). It referenced this issue without a closing keyword, so it stayed open.

josephfusco:
Thanks @jigneshbhavani!

PR #447, opened 2026-09-01
docs: point post-lock-bridge at the block-editor sync split 

theaminulai:
### Why
[#414](https://github.com/WordPress/presence-api/issues/414) flagged that `post-lock-bridge.php`'s header describes the block editor consuming presence data directly as a *future* intent. That's no longer true — it's settled now. The block editor already consumes presence through `WP_Sync_Storage` behind `__unstable_wp_sync_storage` ([WordPress/gutenberg#81697](https://github.com/WordPress/gutenberg/pull/81697)), and [WordPress/sync-storage](https://github.com/WordPress/sync-storage/blob/main/lib/rtc/class-sync-storage-provider.php) delegates `get_awareness_state()`/`set_awareness_state()` straight to `wp_get_presence()`/`wp_set_presence()`. Nothing in this repo documented that split, leaving the header as the only (and outdated) explanation a reader would find.

### What
- Added a "Relationship to the block editor" section to `README.md` explaining the split: editor awareness/cursors flow through sync-storage (CRDT content kept separately in `wp_collaboration`), while room membership across the rest of wp-admin — plus the post-lock bridge — stays this plugin's responsibility. Notes that no room-string mapping is needed since sync-storage's `{objectType}:{objectId}` and this plugin's `postType/{type}:{id}` already agree.
- Replaced the stale "intent is for the future" paragraph in `post-lock-bridge.php`'s file header with a pointer to that new README section.

### How
Doc-only change — no PHP logic, hooks, or public API touched. `wp_presence_bridge_post_lock()` behavior is unchanged.

### Test instructions
1. `git diff` confirms only `README.md` and `includes/post-lock-bridge.php` changed, both comments/docs.
2. Read through the new README section and the updated file header for accuracy against the linked sync-storage source (`class-sync-storage-provider.php`) and gutenberg#81697.
3. Run the existing test suite to confirm no regressions (should be a no-op since no code changed):
  ```
   npm install
   composer install
   composer test
  ```
4. No manual/browser testing needed since this doesn't touch runtime behavior.

**Fixes**: #414

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,7 +205,7 @@ wp presence recording set off --network   # Multisite only
 
 ## Relationship to the block editor
 
-Real-time awareness inside the editor — cursors, selections, who's editing which block — is not this plugin's job. The block editor gets that from [`WP_Sync_Storage`](https://github.com/WordPress/sync-storage), consumed through Gutenberg's `__unstable_wp_sync_storage` ([WordPress/gutenberg#81697](https://github.com/WordPress/gutenberg/pull/81697)). Its [HTTP-polling provider](https://github.com/WordPress/sync-storage/blob/main/lib/rtc/class-sync-storage-provider.php) delegates `get_awareness_state()` and `set_awareness_state()` straight to `wp_get_presence()` and `wp_set_presence()`, while keeping CRDT document updates in its own `wp_collaboration` table.
+Real-time awareness inside the editor — cursors, selections, who's editing which block — is not this plugin's job. The block editor gets that from the [sync-storage](https://github.com/WordPress/sync-storage) plugin, registered through Gutenberg's [`__unstable_wp_sync_storage`](https://github.com/WordPress/gutenberg/pull/81697) filter as an implementation of Gutenberg's `WP_Sync_Storage` interface. Its [storage provider](https://github.com/WordPress/sync-storage/blob/main/lib/rtc/class-sync-storage-provider.php) backs `get_awareness_state()` and `set_awareness_state()` with `wp_get_presence()` and `wp_set_presence()`, while keeping CRDT document updates in its own `wp_collaboration` table.
 
 No room mapping sits between the two: sync-storage's room string (`{objectType}:{objectId}`) and this plugin's `postType/{type}:{id}` room ([`wp_presence_post_room()`](#php-api)) already agree, since `objectType` there is `postType/{type}`.
 
--- a/README.md
+++ b/README.md
@@ -207,7 +207,7 @@ wp presence recording set off --network   # Multisite only
 
 Real-time awareness inside the editor — cursors, selections, who's editing which block — is not this plugin's job. The block editor gets that from the [sync-storage](https://github.com/WordPress/sync-storage) plugin, registered through Gutenberg's [`__unstable_wp_sync_storage`](https://github.com/WordPress/gutenberg/pull/81697) filter as an implementation of Gutenberg's `WP_Sync_Storage` interface. Its [storage provider](https://github.com/WordPress/sync-storage/blob/main/lib/rtc/class-sync-storage-provider.php) backs `get_awareness_state()` and `set_awareness_state()` with `wp_get_presence()` and `wp_set_presence()`, while keeping CRDT document updates in its own `wp_collaboration` table.
 
-No room mapping sits between the two: sync-storage's room string (`{objectType}:{objectId}`) and this plugin's `postType/{type}:{id}` room ([`wp_presence_post_room()`](#php-api)) already agree, since `objectType` there is `postType/{type}`.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -187,6 +187,14 @@ wp presence recording set off
 wp presence recording set off --network   # Multisite only
 ```
 
+## Relationship to the block editor
+
+Real-time awareness inside the editor — cursors, selections, who's editing which block — is not this plugin's job. The block editor gets that from [`WP_Sync_Storage`](https://github.com/WordPress/sync-storage), consumed through Gutenberg's `__unstable_wp_sync_storage` ([WordPress/gutenberg#81697](https://github.com/WordPress/gutenberg/pull/81697)). Its [HTTP-polling provider](https://github.com/WordPress/sync-storage/blob/main/lib/rtc/class-sync-storage-provider.php) delegates `get_awareness_state()` and `set_awareness_state()` straight to `wp_get_presence()` and `wp_set_presence()`, while keeping CRDT document updates in its own `wp_collaboration` table.
```
josephfusco: Three things to untangle here.

`WP_Sync_Storage` is Gutenberg's interface, not sync-storage. The provider's own docblock says so: "Implements WP_Sync_Storage interface from Gutenberg." What the editor gets awareness from is the sync-storage plugin, which implements it.

"Its HTTP-polling provider" merges two separate pieces. The file you link is the storage provider; the HTTP polling server is Gutenberg's, at `lib/experimental/collaboration/class-wp-http-polling-sync-server.php`.

And it does not delegate straight. `get_awareness_state()` reads back only its own rows:

```php
private const CLIENT_PREFIX = 'sync-';
...
if ( 0 !== strpos( $entry->client_id, self::CLIENT_PREFIX ) ) {
    continue;
}
```

The docblock gives the reason: this plugin's heartbeat writes `editor-{user_id}` into the same room, and returning those rows to Gutenberg as awareness "makes the editor throw on a field it has no equality check for."

```suggestion
Real-time awareness inside the editor — cursors, selections, who's editing which block — is not this plugin's job. The block editor gets that from the [sync-storage](https://github.com/WordPress/sync-storage) plugin, registered through Gutenberg's [`__unstable_wp_sync_storage`](https://github.com/WordPress/gutenberg/pull/81697) filter as an implementation of Gutenberg's `WP_Sync_Storage` interface. Its [storage provider](https://github.com/WordPress/sync-storage/blob/main/lib/rtc/class-sync-storage-provider.php) backs `get_awareness_state()` and `set_awareness_state()` with `wp_get_presence()` and `wp_set_presence()`, while keeping CRDT document updates in its own `wp_collaboration` table.
```

README.md
```diff
@@ -187,6 +187,14 @@ wp presence recording set off
...
+
+No room mapping sits between the two: sync-storage's room string (`{objectType}:{objectId}`) and this plugin's `postType/{type}:{id}` room ([`wp_presence_post_room()`](#php-api)) already agree, since `objectType` there is `postType/{type}`.
```
josephfusco: The conclusion is right, the vocabulary isn't. Gutenberg's `WP_Sync_Config::parse_room()` returns `entity_kind`, `entity_name` and `object_id`, so the grammar is `{entity_kind}/{entity_name}:{object_id}`, and it is defined in Gutenberg rather than in sync-storage.

This is also the place for the fact the section is circling: both write into the same room, and the `client_id` prefix is what keeps them apart.

```suggestion
No room mapping sits between the two: both use `postType/{type}:{id}`, the grammar Gutenberg's `WP_Sync_Config::parse_room()` defines and [`wp_presence_post_room()`](#php-api) already returns. What keeps the two sets of rows apart inside that shared room is the `client_id` prefix: sync-storage writes `sync-{id}` and reads nothing else back, leaving this plugin's `editor-{user_id}` rows untouched.
```

includes/post-lock-bridge.php
```diff
@@ -4,10 +4,10 @@
...
  * This bridge is transitional. It records the lock on the editing user's
  * presence entry alongside the existing _edit_lock postmeta so both systems
- * coexist. The intent is for the block editor (Gutenberg) to consume presence
- * data directly in the future — enabling real-time awareness (cursors,
- * selections, who's editing which block) rather than the current blunt
- * lock/takeover model.
+ * coexist. See "Relationship to the block editor" in README.md for how this
+ * relates to the block editor's own awareness data (cursors, selections,
+ * who's editing which block), which flows through WP_Sync_Storage rather
+ * than through this bridge.
```
josephfusco: Same naming point as the README: the interface is Gutenberg's, the thing the data flows through is the plugin.

```suggestion
 * who's editing which block), which flows through the sync-storage plugin
 * rather than through this bridge.
```

josephfusco (commented):
Useful section to have, and the hard parts hold up: gutenberg#81697 merged on 17 August, the provider does back awareness with `wp_get_presence()` and `wp_set_presence()`, and the `wp_collaboration` line is verbatim from the provider's own file header. The corrections below are all attribution, from reading `WordPress/sync-storage@main` and `WordPress/gutenberg@trunk`.

PR #448, opened 2026-09-01
fix: distinguish a non-aggregating network from a quiet one

### LINKED ISSUE #411: Network Admin screens report nobody online when aggregation is switched off
`wp_presence_network_aggregation_enabled()` gates the reads as well as the push, and the read returns the empty summary. A network that has stopped aggregating therefore reports that nobody is online anywhere, rather than reporting that it does not aggregate.

## Evidence

Two callers on `main`, the push and the read:

```console
$ git grep -n "wp_presence_network_aggregation_enabled() ||" -- includes/
includes/network-functions.php:210:	// wp_presence_push_network_summary()
includes/network-functions.php:818:	// wp_presence_compute_network_snapshot()
```

`wp_presence_compute_network_snapshot()` returns `wp_presence_empty_network_summary()` when the gate is closed, and every surface reads through it:

| Surface | Reads at | Via |
| --- | --- | --- |
| Network dashboard Who's Online | `includes/widgets/class-wp-presence-network-widget-whos-online.php:106` | `wp_presence_get_network_summary()` |
| Network Sites list, Online column | `includes/network-sites-list.php:67` | `wp_presence_get_network_summary()` |
| Network Users list, Online view and column | `includes/network-user-list.php:26`, `:67` | `wp_presence_get_network_online_user_ids()` |
| `GET /wp-presence/v1/presence/network` | `includes/class-wp-rest-presence-network-controller.php:144`, `:191` | `wp_presence_get_network_summary()` |
| `wp presence network` | `includes/cli/class-wp-presence-cli-command.php:239` | `wp_presence_get_network_summary()` |

The gating is deliberate and documented: drawing rows nobody is refreshing would be worse. The gap is that "empty" is the only thing the read can say.

## Reproduction

On a multisite install with users online across several sites:

```php
add_filter( 'wp_presence_network_aggregation_enabled', '__return_false' );
```

Load Network Admin > Dashboard. The widget is empty. The Sites list Online column reads `0` for every site. Nothing distinguishes this from a quiet network.

## Worth knowing

A network crosses this on its own. The default is `! wp_is_large_network()`, so passing 10,000 sites flips it, and the screens go from correct to confidently wrong with nothing said.

`0` and "nobody is online" are answers. "This network does not aggregate presence" is the true one, and the screens have no way to say it.

## Tasks

- [x] Give the network read a way to distinguish "not aggregating" from "empty"
- [x] Say so on the surfaces, at least the dashboard widget and the Sites list column

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

Used for: Assisting with the audit of the network read path

</details>

i-am-chitti:
Fixes #411

The network read now returns an `aggregating` flag, so an empty result can say which kind of empty it is. Only the switch sets it false - a network still waiting on its summary table aggregates and has nothing yet.

- **Dashboard widget** - says so instead of "No users are currently online anywhere on the network." The flag is in the Heartbeat hash, or a quiet network and a gated one send an identical payload and the tick reports `unchanged` forever.
- **Sites list** - one `notice-warning` above the table rather than a sentence down every row; cells keep their em dash.
- **REST** - `X-WP-Presence-Aggregating` on both routes.
- **WP-CLI** - `wp presence network` warns instead of printing `Sites online: 0`. `--format=json` already carries the flag.

### Screenshots

| Aggregating | Not aggregating |
| --- | --- |
| widget-aggregating | widget-not-aggregating |

sites-list-not-aggregating

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5
Used for: Auditing the network read path, then implementing the change and its tests.

</details>

### DIFF
--- a/includes/class-wp-rest-presence-network-controller.php
+++ b/includes/class-wp-rest-presence-network-controller.php
@@ -162,6 +162,9 @@ class WP_REST_Presence_Network_Controller extends WP_REST_Controller {
 		$response->header( 'X-WP-Total', $summary['total_sites_online'] );
 		$response->header( 'X-WP-TotalPages', (int) ceil( $summary['total_sites_online'] / $per_page ) );
 		$response->header( 'X-WP-Presence-Users-Online', $summary['total_users_online'] );
+		// An empty collection is two answers, and the body and status are the
+		// same for both.
+		$response->header( 'X-WP-Presence-Aggregating', $summary['aggregating'] ? '1' : '0' );
 		$response->header( 'Cache-Control', 'no-store' );
 
 		return $response;
@@ -199,6 +202,7 @@ class WP_REST_Presence_Network_Controller extends WP_REST_Controller {
 
 		$response = rest_ensure_response( $this->prepare_item_for_response( $item, $request )->get_data() );
 
+		$response->header( 'X-WP-Presence-Aggregating', $summary['aggregating'] ? '1' : '0' );
 		$response->header( 'Cache-Control', 'no-store' );
 
 		return $response;
--- a/includes/cli/class-wp-presence-cli-command.php
+++ b/includes/cli/class-wp-presence-cli-command.php
@@ -255,6 +255,13 @@ class WP_Presence_CLI_Command extends WP_CLI_Command {
 			return;
 		}
 
+		// Reported instead of the totals, not alongside them: zero sites and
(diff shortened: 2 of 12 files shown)


josephfusco (approved):
Nice work!

Issue #132, opened 2026-07-26
Produce a scale analysis and benchmark before the core proposal


josephfusco:
The current write behavior -- unconditional upsert on every heartbeat tick per user -- has not been benchmarked or analyzed for impact at realistic concurrency levels.

Before this becomes a core proposal, there should be a documented answer to: what does the `wp_presence` table look like under 100, 500, and 1000 concurrent admin users? What are the query times, lock wait times, and table size at those levels with the 60-second TTL and the per-minute cleanup cron?

This does not require a solution to the write volume question -- it requires evidence. A tested environment with realistic data that shows either the current design handles expected load acceptably, or that specific numbers justify the engineering changes tracked in #131.

Without this, the core proposal will face the question at review time without a prepared answer.

josephfusco:
This issue was identified as part of a peer review conducted by Claude Sonnet 4.6, facilitated by @josephfusco.

jigneshbhavani:
I'd like to take this one.

One thing worth settling before picking numbers to chase: the table is bounded by concurrency, not by elapsed time. Each user holds one `admin/online` row plus one per post editor they have open, and cleanup drops anything past the 60s TTL every minute (`includes/functions.php:456` and `:467` cap it at 1000 rows times 10 passes). So 1000 concurrent admins is on the order of 1000 to 2000 rows, which is small. If that holds, table size isn't where the answer is, and the numbers that matter are write contention on one hot table and what each tick costs to read.

On method, since the issue doesn't specify: I'd drive the real heartbeat endpoint with N concurrent workers rather than simulating the writes, so the per-tick read path is included, and report query and lock wait times at 100 / 500 / 1000 with the cleanup cron running against them.

Flagging one overlap: #205 covers the uncached aggregates on every tick and is already assigned. If the read side turns out to be the headline here, that's its territory rather than mine, so I'll keep to measuring and hand over anything it should own.

jigneshbhavani:
Benchmarked it. Writes and table size are fine. The problem is the heartbeat response.

**Cause.** The Who's Online handler runs on every tick from every user, not just people on the Dashboard (`presence-api.php:301`, guarded only on `presence-ping`, which every page sends). Each tick it reads the whole `admin/online` room, looks up every user and avatar, and sends the lot back. The widget then draws 3 rows (`VISIBLE_ROWS`, sliced at `:624`). The slice happens after the send, not before, so nearly all of it is thrown away.

[...]

PR #450, opened 2026-09-01
perf: skip presence writes that would only move the timestamp

### LINKED ISSUE #131: Add state-change detection before upsert to reduce write volume

jigneshbhavani:
Presence is upserted on every Heartbeat tick, so a user sitting on one screen rewrites the same row with byte-identical data until they leave. This compares the encoded state against the stored row first and leaves it alone when nothing changed and the row will still be inside the cutoff at the next tick.

Fixes #131

One live editor session, same steps on both branches:

| | Writes in 186s | Max row age while live |
| --- | --- | --- |
| `main` | one every ~10s, payload hash unchanged every time | n/a, refreshed constantly |
| This branch | 3, each a real state change | 95s against the 150s TTL |

The threshold is `wp_presence_get_timeout() - 15 - gap`. It derives from the filtered timeout rather than `WP_PRESENCE_DEFAULT_TTL`, or a site filtering the TTL below the tick interval blinks its users offline. The 15 mirrors `TTL_SAFETY_MARGIN` in `presence-ping.js`, which caps the client's own idle backoff against the same TTL.

`gap` comes from `has_focus`, not from the `interval` in the same request. `scheduleNextTick()` hard-codes `120000` on a blurred window and never writes that back into the reported field, so the reported value understates the real gap on exactly the case with the least headroom: a 150s TTL against a 120s gap.

**The multisite coupling is the part worth reviewing.** The network summary push runs off `wp_presence_admin_room_changed` (`presence-api.php:329`), which only fires on a real write. Skipping admin-room writes decays the summary row past the read cutoff and the Network Admin screens report nobody online. `wp_presence_network_summary_push_is_due()` keeps those writes.

Notes:

- Only the staleness half of `wp_presence_network_summary_needs_push()` applies to that guard. A membership change is somebody else's state changing, and their own write announces it.
- The cost is one `SELECT` on the `(room, client_id)` unique key before each write. This does not contradict #132, which measured the upsert as cheap and it still is. What this removes is write volume, not tick latency.
- I left the threshold unfiltered. Adding one trips the public surface workflow and is a commitment to keep, which is the argument @mindctrl made on #437.
- Not run on SQLite. Everything below is MariaDB.

Testing:

- `phpunit` — 432 tests, 719 assertions, 108 skipped
- `WP_MULTISITE=1 phpunit` — 432 tests, 928 assertions, 3 expected skips
- `./vendor/bin/phpcs --standard=phpcs.xml.dist` — passed
- `./vendor/bin/phpstan analyse --no-progress` — passed
- Browser A/B on wp-env against both branches, sampling `wp_presence` through the session
- Four mutation checks, each caught by the intended test: removing the guard, making the threshold never expire, dropping the state comparison, removing the network bail-out

<details>

AI assistance: Yes  
Tool(s): Claude Code  
Model(s): Claude Opus 5  
Used for: Assisting with the implementation, tests, and this description. The change and the results above were reviewed by me before submission.

</details>

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -187,7 +187,7 @@ function wp_presence_recording_enabled() {
  *
  * @access private
  *
- * @since 0.3.0
+ * @since 0.4.0
  *
  * @return int Seconds.
  */
@@ -221,7 +221,7 @@ function wp_presence_next_tick_gap() {
  *
  * @access private
  *
- * @since 0.3.0
+ * @since 0.4.0
  *
  * @return int Age in seconds. 0 means never skip.
  */
@@ -240,7 +240,7 @@ function wp_presence_refresh_threshold() {
  *
  * @access private
  *
- * @since 0.3.0
+ * @since 0.4.0
  *
  * @param string $room      The room identifier.
  * @param string $client_id The client identifier.
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -183,6 +183,106 @@ function wp_presence_recording_enabled() {
 }
 
 /**
+ * Returns how long until the client is expected to send its next Heartbeat.
+ *
+ * @access private
+ *
+ * @since 0.3.0
+ *
+ * @return int Seconds.
+ */
+function wp_presence_next_tick_gap() {
+	// Core's scheduleNextTick() overrides the interval to 120s whenever the
+	// window is blurred, and never reflects that back into the interval it
+	// reports in the same request, so the reported value understates the real
+	// gap on an unfocused tab. That is the worst case, and the assumption to
+	// make whenever the request does not say otherwise.
+	$blurred = 120;
+
+	// phpcs:ignore WordPress.Security.NonceVerification.Missing -- Nonce verification is handled by WordPress in wp_ajax_heartbeat() before any of this runs.
+	if ( ! isset( $_POST['interval'], $_POST['has_focus'] ) || 'true' !== $_POST['has_focus'] ) {
+		return $blurred;
+	}
+
+	// phpcs:ignore WordPress.Security.NonceVerification.Missing -- As above.
+	$interval = absint( $_POST['interval'] );
+
+	return $interval > 0 ? $interval : $blurred;
+}
+
+/**
+ * Returns the age at which an unchanged presence row still has to be rewritten.
+ *
+ * Skipping a write leaves the row's existing date_gmt in place, so it is only
+ * safe while the row will still be inside wp_get_presence()'s cutoff when the
+ * next tick arrives. Deriving this from wp_presence_get_timeout() rather than
+ * WP_PRESENCE_DEFAULT_TTL matters: a site filtering the TTL below the tick
+ * interval would otherwise make its users blink offline.
+ *
+ * @access private
+ *
+ * @since 0.3.0
+ *
+ * @return int Age in seconds. 0 means never skip.
+ */
+function wp_presence_refresh_threshold() {
+	// Mirrors TTL_SAFETY_MARGIN in assets/js/presence-ping.js, which caps the
+	// client's own idle backoff against the same TTL. The two have to agree.
+	$margin = 15;
+
+	$timeout = wp_presence_get_timeout( WP_PRESENCE_DEFAULT_TTL );
+
+	return max( 0, $timeout - $margin - wp_presence_next_tick_gap() );
+}
+
+/**
+ * Reports whether a write would leave the stored row exactly as it already is.
+ *
+ * @access private
+ *
+ * @since 0.3.0
+ *
+ * @param string $room      The room identifier.
+ * @param string $client_id The client identifier.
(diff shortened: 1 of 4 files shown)


josephfusco:
The change reads right and the A/B table makes the case well. The coverage is thinner than the change deserves, though. Four mutations survive all six new tests, and each is one test to close.

**`$threshold <= 0`.** Filtering the TTL below the tick interval is why the threshold derives from `wp_presence_get_timeout()`, and it's the one branch nothing reaches. Drop the guard, or turn `max( 0, ... )` into `abs()`, and all six tests still pass. Filter the timeout to 60, write twice unchanged, assert `date_gmt` moved.

**The multisite bail-out.** Delete the `wp_presence_admin_room()` block from `wp_presence_write_is_redundant()` and the suite stays green. The two `push_is_due()` tests exercise that function alone; nothing asserts that a due push keeps the write alive. You list this mutation among the four you ran, so which test caught it? The one I'd expect is multisite, aggregation on, push overdue, unchanged admin-room state, assert the row was written.

**`wp_presence_next_tick_gap()`.** No test sets `$_POST['interval']` or `$_POST['has_focus']`. `$blurred` can be anything from 0 to 134 without a failure, so the value the whole threshold rests on isn't pinned. That branch gets the longest paragraph in your description and no coverage.

**The staleness comparison in `push_is_due()`.** Flip `<=` to `>=` and both tests pass, because the never-pushed one returns at the `! is_array` check before reaching it. Set `wp_presence_network_pushed` to a recent time, assert false.

Last thing is a question. `test_skipped_write_does_not_announce_an_admin_room_change` asserts the action never fires. Single site is fine, since `wp_presence_network_aggregation_enabled()` bails on `! is_multisite()` first. Under `WP_MULTISITE=1` with the summary table registered and nothing pushed yet, `push_is_due()` returns true, the write goes through and the action fires. Your multisite run is green, so what's keeping it false there?

Also, 0.3.0 is going out without this one, so those four `@since 0.3.0` tags want to be `0.4.0`.

jigneshbhavani:
@josephfusco Good catch on all four, all in `f1cf689`, in your order:

- `$threshold <= 0`: `test_a_ttl_below_the_tick_gap_leaves_no_room_to_skip`, plus `test_a_zero_threshold_writes_again_inside_the_same_second` for the reason at the bottom
- the bail-out: `test_an_overdue_push_keeps_an_unchanged_admin_room_write`
- `wp_presence_next_tick_gap()`: `test_the_tick_gap_assumes_a_blurred_window_unless_the_request_says_otherwise`
- the staleness comparison: `test_push_is_not_due_when_the_row_was_pushed_recently`

On the bail-out, there is a pre-existing test that goes red when the block is deleted, `test_push_refreshes_a_stale_row_when_nobody_changed`, but it is not one of mine and not one I want to lean on, since it was written for something else and could reasonably change.

Your last question has an answer I did not enjoy. `WP_Presence_UnitTestCase` never provisions the summary table, so `has_network_summary_table()` is false there and `push_is_due()` returns false whatever `WP_MULTISITE` is set to. That test was passing for the environment rather than the logic, and I would not have gone looking without you asking. `ff209f3` asserts the precondition so it stops reading as coverage of a case it never touched.

The zero threshold took two tests and I am not convinced the second earns its place. Yours catches `abs()`. But keep `max( 0, ... )`, drop only the early return, and `age <= 0` is already false at any age above zero, so the row gets written and the test passes anyway. The only case that early return decides on its own is a second write inside the same second, which is why that one counts INSERTs instead. Is the branch worth keeping, or should the comparison just carry it?

`@since` is on 0.4.0, thanks, I would have shipped that wrong.

josephfusco (approved):
LGTM, nice addition!

PR #453, opened 2026-09-02
fix: centralize network site status filtering

### LINKED ISSUE #398: Only one of the three network site-status filters is reachable
The network read path filters archived, spam, and deleted sites at three `get_sites()` calls. Only `wp_presence_compute_network_snapshot()` is positioned to drop anything, because the other two read its output.

`wp_presence_hydrate_network_snapshot()` has one caller ([`network-functions.php:535`](../blob/main/includes/network-functions.php#L535)), and the `sites-by-user` index is built at [`:409`](../blob/main/includes/network-functions.php#L409). Both take compute's return value, and the `presence_network` cache group is non-persistent, so a snapshot cannot outlive the request that built it.

The `is_array()` guards alongside them are likewise unreachable: `get_sites()` returns an array unless `'count' => true` is passed, which nothing here does.

## Evidence

One edit each, then `WP_MULTISITE=1 ./vendor/bin/phpunit --group ms-required`:

| Mutation | Result |
| --- | --- |
| Drop the six downstream arguments | 62 pass |
| Drop all nine | 3 failures, one per regression test |

## Tasks

- [ ] Narrow the status filter to `wp_presence_compute_network_snapshot()`
- [ ] Drop the five `is_array()` guards
- [ ] Confirm each regression test fails when its flag is removed from compute, now that nothing masks it

Leave `uninstall.php` and `wp_presence_get_network_site_ids()` unfiltered by status. Both need to reach archived, spam, and deleted sites, or their presence tables are left behind.

Follow-up to #351.

<details>

- AI assistance: Yes
- Tool(s): Claude Code
- Model(s): Opus 5
- Used for: Assisting with mutation testing and drafting this issue.

</details>

Junaid-PK:
The network snapshot already removes archived, spam, and deleted sites before downstream callers receive it. This change keeps the status filter at that single source of truth, removes the duplicate arguments from the user-site lookup and hydration queries, and removes the three unreachable `is_array()` guards around `get_sites()` results.

The two other `is_array()` checks in this file remain because they validate option and decoded JSON data rather than `get_sites()` return values.

Joe Fusco is credited as a co-author for the mutation-tested diagnosis and implementation design in the issue.

Fixes #398

## Testing

- `php -l includes/network-functions.php`
- `./vendor/bin/phpcs --standard=phpcs.xml.dist includes/network-functions.php`
- `./vendor/bin/phpstan analyse --configuration=phpstan.neon.dist --memory-limit=2G --no-progress`

The multisite PHPUnit group could not be run locally because a Docker daemon was unavailable; the existing CI matrix covers it.

<details>

AI assistance: Yes
Tool(s): OpenAI coding assistant
Model(s): GPT-5
Used for: Implementing the focused change, rebasing it onto current `main`, and running the static checks listed above.

</details>


josephfusco:
Hey @Junaid-PK, nice first contribution! 

When you get a moment, please connect a WordPress.org account so that you get proper attribution once this merges. [See comment](https://github.com/WordPress/presence-api/pull/453#issuecomment-5510272322)

Junaid-PK:
Thanks for the reminder, Joe. I’ll link my WordPress.org profile before the PR is merged so the attribution is picked up correctly.

josephfusco (approved):
LGTM! Drop a comment here once your .org account is linked up so we can run the props-bot again and get this merged in.

Issue #451, opened 2026-09-02
Collaboration edge is only exposed to PHP, so the client cannot act on it


josephfusco:
`wp_presence_check_collaboration_threshold()` fires `wp_presence_collaboration_started` and `_ended` on the 1 to 2+ and 2+ to 1 edges. Both are PHP actions. The editor heartbeat calls it at [heartbeat.php:334](https://github.com/WordPress/presence-api/blob/main/includes/heartbeat.php#L334) and then returns `$response` unchanged, so a browser has no way to learn the edge without polling something.

That makes the signal unusable for the case it exists for: letting the block editor decide whether to start the sync loop. Gutenberg derives `hasCollaborators` from its own poll response today, so a solo editor pays ~840 requests an hour to discover nobody else is there.

## Done when

- [x] The editor heartbeat response carries the current editor count for the room
- [x] `wp_presence_check_collaboration_threshold()` returns the count instead of void
- [x] The new key is in `RELAYED_TICK_KEYS` (`assets/js/presence-ping.js`), or a coalesced sibling tab never receives it
- [x] No filter on the key. None of the existing response keys has one, and #450 records the argument against adding filters that become commitments
- [x] Tests cover the key's presence on the editor screen and its absence elsewhere

## Shape

| | |
| --- | --- |
| Handler | `wp_presence_editor_heartbeat_received()`, `includes/heartbeat.php:288` |
| Precedent for setting a key | `includes/debugger-widget.php:115` |
| Proposed key | `presence-heartbeat-collaborators` |
| Return change | `wp_presence_check_collaboration_threshold()` void to int, back compatible |

Part of #444, section 6. Prerequisite for deferring the Gutenberg poll loop until a second person arrives.

<details>
<summary>Use of AI Tools</summary>

AI assistance: Yes
Tool(s): Claude Code
Used for: Assisting with drafting this issue.
</details>

josephfusco:
Hey @theaminulai, please comment on the issue you want so a maintainer can assign it. GitHub only allows assignment to people who have commented. This helps avoid duplicating work.

https://github.com/WordPress/presence-api/blob/main/.github/CONTRIBUTING.md#claiming-an-issue

theaminulai:
I’d like to work on this issue. Please assign it to me. @josephfusco  Thanks!

josephfusco:
Closing — #454 already covered all five items above.

PR #454, opened 2026-09-02
feat: carry the room's editor count on the editor heartbeat response 

theaminulai:
## Why
[#451](https://github.com/WordPress/presence-api/issues/451)  `wp_presence_check_collaboration_threshold()` already fires `wp_presence_collaboration_started`/`_ended` PHP actions on the 1↔2+ editor edge, but those are server-side only. The editor heartbeat handler calls it and discards the result, so a browser has no way to learn the edge without polling something separately. That makes the signal unusable for the case it exists for: letting the block editor decide whether to start its real-time sync loop. Today Gutenberg derives `hasCollaborators` from its own poll response, costing a solo editor ~840 requests an hour to discover nobody else is there.

## What
- `wp_presence_check_collaboration_threshold()` now returns the current editor count (int) instead of void, back-compatible, since nothing previously used the return value.
- `wp_presence_editor_heartbeat_received()` sets that count on the response as `presence-heartbeat-collaborators`.
- `presence-heartbeat-collaborators` added to `RELAYED_TICK_KEYS` in `assets/js/presence-ping.js` so a coalesced sibling tab still receives it.
- No filter added on the key, consistent with every other `presence-heartbeat-*` key, and with the argument recorded in #450 against filters that become commitments.

## How
- `includes/heartbeat.php`: `$editor_count` is computed and used for the started/ended action edge exactly as before; the only change is returning it, and assigning it into `$response` at the call site.
- `assets/js/presence-ping.js`: one-line addition to the existing relay list, same pattern as the other `presence-heartbeat-*` keys.
- No new hooks, no new REST surface, no schema change.

## Test instructions
1. `composer phpcs` / `composer phpstan` both clean on the changed files.
2. `npm test` (needs `wp-env`/Docker)  runs the PHPUnit suite, including new/updated tests in `tests/test-heartbeat.php`:
   - `test_editor_heartbeat_writes_presence`  updated to assert the full response shape now includes `presence-heartbeat-collaborators`.
   - `test_editor_heartbeat_response_carries_the_room_s_editor_count`  two editors ticking → response reports `2`.
   - `test_editor_heartbeat_omits_collaborator_count_without_a_post_id`  a ping that never reaches the write path carries no key.
   - `test_admin_heartbeat_omits_collaborator_count`: the unrelated admin heartbeat response never carries the key.
3. Manual sanity check: open a post in two browser sessions as different editors, watch Heartbeat responses in devtools; `data.presence-heartbeat-collaborators` should read `2` on both, and drop to `1` when one leaves.

### DIFF
--- a/includes/heartbeat.php
+++ b/includes/heartbeat.php
@@ -362,7 +362,7 @@ function wp_presence_collaboration_state_key( $room ) {
  * Fires 'wp_presence_collaboration_ended' when editor count goes from 2+ to 1.
  *
  * @since 0.1.21
- * @since 0.3.0 Returns the current editor count instead of void.
+ * @since 0.4.0 Returns the current editor count instead of void.
  *
  * @param string $room The presence room identifier.
  * @return int The number of editors currently present in the room.
--- a/includes/heartbeat.php
+++ b/includes/heartbeat.php
@@ -280,6 +280,10 @@ function wp_presence_admin_heartbeat_received( $response, $data, $screen_id ) { 
 /**
  * Handles the editor presence heartbeat and creates a presence entry for the post being edited.
  *
+ * Also carries the room's current editor count back as
+ * `presence-heartbeat-collaborators`, so a client can decide whether to start
+ * a real-time sync loop without polling for it separately.
+ *
  * @param array  $response  The Heartbeat response.
  * @param array  $data      The $_POST data sent.
  * @param string $screen_id The screen ID.
@@ -331,7 +335,7 @@ function wp_presence_editor_heartbeat_received( $response, $data, $screen_id ) {
 		$user_id
 	);
 
-	wp_presence_check_collaboration_threshold( $room );
+	$response['presence-heartbeat-collaborators'] = wp_presence_check_collaboration_threshold( $room );
 
 	return $response;
 }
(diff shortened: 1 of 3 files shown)


josephfusco (dismissed):
`@since 0.3.0` on `wp_presence_check_collaboration_threshold()` needs to be `0.4.0`. 0.3.0 shipped on 2 September. release-please runs with `bump-minor-pre-major`, so a `feat:` lands as 0.4.0; `includes/functions.php` on main is already tagged that way.

The rest holds up. The threshold function has a single exit, so the response key is always an int. `wp_get_presence()` applies no LIMIT, which keeps the count honest in a busy room. `pingContextKey` already carries `editorPostId`, so a relayed key never reaches a tab on a different post. phpcs, phpstan and the full single-site suite pass. The multisite failures on my machine happen on main too.

Not a blocker: no heartbeat response key is documented anywhere today, but this one is meant for Gutenberg to read. README's "Relationship to the block editor" section seems like the place for a line about it, here or as a follow-up.

Thanks for picking this up.

includes/heartbeat.php
```diff
@@ -358,8 +362,10 @@ function wp_presence_collaboration_state_key( $room ) {
  * Fires 'wp_presence_collaboration_ended' when editor count goes from 2+ to 1.
  *
  * @since 0.1.21
+ * @since 0.3.0 Returns the current editor count instead of void.
```
josephfusco:
```suggestion
 * @since 0.4.0 Returns the current editor count instead of void.
```

josephfusco (approved):
LGTM!

Issue #452, opened 2026-09-02
wp_set_presence() always stamps date_gmt with now, so a relay cannot preserve the original time


josephfusco:
`wp_set_presence( $room, $client_id, $state, $user_id = 0 )` hardcodes `$now = gmdate( 'Y-m-d H:i:s' )` and writes that as `date_gmt` ([functions.php:203](https://github.com/WordPress/presence-api/blob/main/includes/functions.php#L203)).

A caller relaying awareness on behalf of other clients therefore cannot preserve their timestamps. Every relayed entry is stamped with the relay's own clock, so whoever is polling holds everyone else in the room alive. sync-storage hit this: before [sync-storage#91](https://github.com/WordPress/sync-storage/pull/91) each client rewrote every collaborator's row on every poll, which kept departed collaborators in the room.

## Done when

- [ ] `wp_set_presence()` accepts an optional GMT timestamp
- [ ] A supplied timestamp in the future is clamped to now, or a caller can pin a row past the TTL indefinitely
- [ ] The skip guard from #450 does not swallow a write that carries an explicit timestamp
- [ ] `@since` and the README's PHP API section updated
- [ ] Tests cover the default, an explicit past timestamp, and a future one

## Conflict to resolve first

#450 adds a guard that reads the stored `date_gmt` and skips the write when state is unchanged and the row stays inside the cutoff. An explicit timestamp has to bypass it. Worth landing #450 first and building this on top.

Part of #444, section 6.

<details>
<summary>Use of AI Tools</summary>

AI assistance: Yes
Tool(s): Claude Code
Used for: Assisting with drafting this issue.
</details>

theaminulai:
I’d like to work on this issue. Please assign it to me. @josephfusco Thanks!

josephfusco:
Closed by #456

PR #456, opened 2026-09-02
feat: let wp_set_presence() accept an explicit GMT timestamp

theaminulai:
## Why
[#452](https://github.com/WordPress/presence-api/issues/452) `wp_set_presence()` hardcodes `$now = gmdate( 'Y-m-d H:i:s' )` as the row's `date_gmt`, so a caller relaying awareness on behalf of other clients (like sync-storage bridging Gutenberg collaborators) can't preserve their real timestamps. Every relayed entry gets stamped with the relay's own clock, so whoever happens to be polling holds everyone else in the room alive. sync-storage hit exactly this before [sync-storage#91](https://github.com/WordPress/sync-storage/pull/91): each client rewrote every collaborator's row on every poll, keeping departed collaborators in the room.

## What
- `wp_set_presence()` gains an optional 5th parameter, `$date_gmt` (`'Y-m-d H:i:s'`, same shape `wp_get_presence()` already returns), defaulting to `null` (current behavior: now).
- A supplied timestamp in the future is clamped to now via `min( $date_gmt, $current )`, so a caller can't pin a row past the TTL indefinitely.
- `@since 0.3.0` added on the parameter; README's PHP API snippet for `wp_set_presence()` updated to document it.

## How
`$now` is computed as `null === $date_gmt ? $current : min( $date_gmt, $current )` string comparison works because `'Y-m-d H:i:s'` sorts lexicographically the same as chronologically. Everything downstream (the `INSERT ... ON DUPLICATE KEY UPDATE`) is unchanged; `$now` just has a new source.

**Scope note:** #452 also lists "the skip guard from #450 does not swallow a write that carries an explicit timestamp" as done-when criteria. #450 (a guard that skips a write when state is unchanged and the row is inside the cutoff) hasn't landed in this repo yet; there's currently no such guard in `wp_set_presence()` to interact with. This PR implements #452 standalone; whoever lands #450 will need to make sure its "skip" branch checks for `null !== $date_gmt` (or equivalent) before discarding an explicitly-timestamped write.

## Test instructions
1. `composer phpcs` / `composer phpstan`  both clean.
2. `npm test` (needs `wp-env`/Docker); new tests in `tests/test-functions.php`:
   - `test_set_presence_defaults_date_gmt_to_now`  no `$date_gmt` arg still stamps now.
   - `test_set_presence_accepts_an_explicit_past_timestamp` a past timestamp is stored verbatim.
   - `test_set_presence_clamps_a_future_timestamp_to_now`  a future timestamp is clamped, not trusted.
3. Manual: call `wp_set_presence( $room, 'client-1', [], 0, gmdate('Y-m-d H:i:s', time() - 300) )` via `wp eval` or a custom mu-plugin, then `wp presence list` and confirm the row's age reflects the past timestamp, not "just now."


includes/functions.php
```diff
@@ -206,7 +218,8 @@ function wp_set_presence( $room, $client_id, $state, $user_id = 0 ) {
 	}
 
 	$data_json = wp_json_encode( $state );
-	$now       = gmdate( 'Y-m-d H:i:s' );
+	$current   = gmdate( 'Y-m-d H:i:s' );
+	$now       = null === $date_gmt ? $current : min( $date_gmt, $current );
```
i-am-chitti, line 358: `$date_gmt` needs validating as a real date before it gets here. Invalid input string can be passed and processed incorrectly. A format check up front would be better - 

```php
if ( null !== $date_gmt && ! DateTimeImmutable::createFromFormat( 'Y-m-d H:i:s', $date_gmt ) ) {
      return false;
}
```

i-am-chitti (commented):
Thanks for the contribution.

#450 has been merged, so the scope note is stale. Once, this PR merges the bug is in `main`.

`wp_presence_write_is_redundant()` never sees `$date_gmt`, so it skips the write. A relay backdating a collaborator who has left gets ignored, and they stay in the room. The parameter does nothing for the case it was added for. 

Fix looks like one line at [includes/functions.php:311](https://github.com/WordPress/presence-api/blob/f8d578d21fac3fa5fe37c8a94ebb91c96b9a3e9c/includes/functions.php#L311), plus a test for that combination:

```php
if ( null === $date_gmt && wp_presence_write_is_redundant( $room, $client_id, $data_json ) ) {
      return true;
}
```

includes/functions.php
theaminulai: Thanks for the feedback. I agree that `$date_gmt` should be validated before it is processed.

One question regarding the validation approach: as far as I can see, `DateTimeImmutable::createFromFormat()` is not currently used in the WordPress codebase. Would you prefer us to use this native PHP API here, or should we follow an existing WordPress date/time validation pattern instead?

I’d prefer to keep this consistent with existing WordPress Core conventions if there is an appropriate utility/pattern available.

josephfusco: Validate it up front with a `preg_match` on the full `Y-m-d H:i:s` plus [`wp_checkdate()`](https://developer.wordpress.org/reference/functions/wp_checkdate/), which is what [`wp_resolve_post_date()`](https://developer.wordpress.org/reference/functions/wp_resolve_post_date/) does.

includes/functions.php
```diff
@@ -288,13 +288,25 @@ function wp_presence_write_is_redundant( $room, $client_id, $data_json ) {
  * Uses INSERT ... ON DUPLICATE KEY UPDATE for atomic upserts
  * via the UNIQUE KEY (room, client_id).
  *
- * @param string $room      The room identifier.
- * @param string $client_id The client identifier.
- * @param array  $state     The presence state data.
- * @param int    $user_id   Optional. The user ID. Default 0.
+ * @since 0.3.0 Added the $date_gmt parameter.
```
josephfusco:
```suggestion
 * @since 0.4.0 Added the $date_gmt parameter.
```

josephfusco (changes requested):
This is nearly there and just needs the guard bypass so an explicit timestamp is not dropped, plus the format check from the thread above.

theaminulai:
Fixed both — thanks for narrowing it down.

`wp_presence_write_is_redundant()` never saw `$date_gmt`, so an explicit timestamp meant for backdating a departed collaborator was silently skipped along with any other unchanged-state write. A relay could no longer report that a collaborator had left.

Bypassed the guard whenever `$date_gmt` is explicit, and validated it up front with a `preg_match` on the literal `'Y-m-d H:i:s'` shape plus `wp_checkdate()` (the same technique `wp_resolve_post_date()` uses internally) so a malformed or impossible date is rejected instead of reaching the database.

Also added test coverage for both: the guard now correctly lets an explicit timestamp through even when the state is unchanged, and a malformed/impossible `$date_gmt` returns `false` without touching the row.

josephfusco (approved):

PR #465, opened 2026-09-02
ci: fail the Plugin Check job when Plugin Check reports an error

### LINKED ISSUE #412: Plugin Check job reports success while printing errors
`wp plugin check` exits 0 whatever it finds, and the workflow sets no severity threshold, so the job reports success while printing errors.

## Evidence

Run 33254997943, `plugin-check` job, 2026-08-29, conclusion `success`:

```
0	0	ERROR	outdated_tested_upto_header	Tested up to: 7.0 < 7.1. The "Tested up to" value in your
plugin is not set to the current version of WordPress. This means your plugin will not show up in
searches, as we require plugins to be compatible and documented as tested up to the most recent
version of WordPress.
```

Nothing runs after that line but the runner's own cleanup. The step passed.

WordPress 7.1 shipped on 2026-08-19. The same error prints and the same job passes on run 33120678715 from 2026-08-27, so this is not a one-off.

```yaml
# .github/workflows/plugin-check.yml:52
- run: wp-env run cli wp plugin check presence-api --format=strict-table
```

`--format=strict-table` selects a renderer. It is not a severity gate, and there is none.

## Tasks

- [ ] Fail the job on `ERROR`, through `--severity` or the step's own exit code
- [ ] Decide whether `WARNING` fails too, given the run above also carries `PluginCheck.Security.DirectDB.UnescapedDBParameter` on the two `dbDelta` call sites
- [ ] Confirm the gate fails by running it against `Tested up to: 7.0`

## Worth knowing

The error message asserts the plugin will not show up in searches. That is not what happens at one major behind: the .org query API returns `presence-api` for `presence api`, and `heartbeat-controller` at `tested 6.8.8` is three majors behind and still returned. The header is worth fixing on its own terms; search removal is not the reason.

The header itself is fixed separately. This issue is the reason nobody saw it.

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

Used for: Assisting with the audit of CI gating

</details>

i-am-chitti:
## Description

Fixes #412

Plugin Check has no flag that can fail the job. `--severity`, `--error-severity` and `--ignore-warnings` only change what gets printed, and `wp plugin check` exits 0 either way. So the check has to happen in the step itself.

Errors fail the job, warnings don't. The current warnings are all pre-existing, so gating them would fail the job on merge day for problems no PR caused.

Tested both ways. Setting `Tested up to` back to 7.0 brings the error back and the job fails. When there is nothing to report, the command prints `[]` and the job passes.

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5
Used for: Checking the Plugin Check CLI for a severity flag and testing the gate both ways.

</details>

### DIFF
--- a/.github/workflows/plugin-check.yml
+++ b/.github/workflows/plugin-check.yml
@@ -49,4 +49,15 @@ jobs:
           wp-env run cli wp plugin activate presence-api
 
       - name: Run Plugin Check
-        run: wp-env run cli wp plugin check presence-api --format=strict-table
+        run: |
+          wp-env run cli wp plugin check presence-api \
+            --format=strict-json \
+            --fields=file,line,column,type,code,message > plugin-check.json
+
+          jq -r '.[] | "\(.type)  \(.file):\(.line):\(.column)  \(.code)\n    \(.message)"' plugin-check.json
+
+          errors="$( jq '[ .[] | select( .type == "ERROR" ) ] | length' plugin-check.json )"
+          if [ "$errors" -gt 0 ]; then
+            echo "Plugin Check reported $errors error(s)."
+            exit 1
+          fi
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,5 @@ build/
 coverage.xml
 .env
 .wp-env.override.json
+plugin-check.json
 .DS_Store


josephfusco (approved):
LGTM, nice add!

PR #466, opened 2026-09-02
test: retry Playwright tests in CI

### LINKED ISSUE #403: e2e specs hit wordpress.org on every WP-CLI call, so one timeout fails the job

muneeb-ashraf:
Add two Playwright retries in CI while preserving zero retries for local runs. This allows transient wordpress.org timeouts to be retried without masking failures during local development.

Fixes #403

### Testing

- Verified both Playwright projects resolve to 0 retries locally and 2 retries in CI.
- `npx playwright test --config tests/e2e/playwright.config.js --list` - 44 tests discovered.
- `npm run test:unit -- --runInBand` - 21 tests passed.
- `npm run test:scripts` - 79 tests passed.
- `git diff --check`

<details>

AI assistance: Yes
Tool(s): OpenAI Codex
Model(s): GPT-5
Used for: Code and tests.

</details>

### DIFF
--- a/tests/e2e/playwright.config.js
+++ b/tests/e2e/playwright.config.js
@@ -44,6 +44,7 @@ const NETWORK_SPECS = /presence-network-[^/]+\.test\.js$/;
 export default defineConfig( {
 	globalSetup: path.resolve( __dirname, 'global-setup.js' ),
 	reporter: process.env.CI ? [ [ 'github' ] ] : [ [ 'list' ] ],
+	retries: process.env.CI ? 2 : 0,
 	workers: 1,
 	timeout: 100_000,
 	reportSlowTests: null,


i-am-chitti (commented):
Tried it locally: 2 retries in CI, 0 on my machine, all 44 tests still found. Also checked the specs start fresh on a retry. So, no new flakiness.

#372 is already merged. So, no blocker here.

LGTM 🚀 

Nit: @muneeb-ashraf  Could you fill in the template's [Use of AI Tools block](https://github.com/WordPress/presence-api/blob/main/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L34-L48) (AI assistance: / Tool(s): / Model(s):)? The current PR description doesn't say which model.

muneeb-ashraf:
Updated the PR description with the complete disclosure block, including the tool, model, and usage. Thanks for the catch.

josephfusco (approved):

PR #467, opened 2026-09-04
fix: gate the network column renderers on the network capability

### LINKED ISSUE #137: Audit multisite behavior before core proposal
0.2.0 added three Network Admin surfaces that read `wp_presence_network_summary`, a network-global table. Before that every read was scoped by table prefix and the site boundary held by construction. It is now a capability check, so the audit has to establish both that the boundary still holds where it should and that the places crossing it deliberately are gated.

## What crosses the boundary

| Surface | Reads | Gated in |
| --- | --- | --- |
| Sites list Online column | `wp_presence_get_network_summary()` | `wp_presence_register_network_sites_column()` (`includes/network-sites-list.php:19`) |
| Users list Online view and column | `wp_presence_get_network_online_user_ids()`, `wp_presence_get_network_sites_for_user()` | `includes/network-user-list.php:22`, `:55`, `:86` |
| Network dashboard widget | `wp_presence_get_network_summary()` | `WP_Presence_Network_Widget_Whos_Online::register()`, `::heartbeat_received()` |

Every gate is `wp_presence_network_capability()`, which defaults to `manage_network` and is filterable (`includes/network-functions.php:411-418`).

## Acceptance criteria

- [ ] A user with `edit_posts` on site A but not site B is denied `admin/online` on site B. `wp_can_access_presence_room()` (`includes/functions.php:232-247`) calls `user_can()`, which is site-context-aware, and `tests/test-functions.php` only exercises it single-site.
- [ ] A presence row written on site A is not readable from site B. `test_table_name_is_scoped_to_the_switched_site()` (`tests/test-table-creation.php:405`) proves the table name switches; nothing proves the read comes back empty.
- [ ] The two render callbacks with no gate of their own, `wp_presence_render_network_users_column()` (`includes/network-user-list.php:108`) and `wp_presence_render_network_sites_column()` (`includes/network-sites-list.php:62`), are either gated or established as unreachable without the registration that precedes them.
- [ ] The `wp_presence_network_capability` docblock states what a filtered value grants. Lowering it to a site-level capability hands every holder presence for sites they are not a member of, and `tests/test-network-rest-controller.php:107` already filters it to `edit_posts`.

## Worth knowing

Each site pushes its own row into the summary table, so the aggregate read never calls `switch_to_blog()` (`tests/test-network-presence.php:118`). The question here is who may read the aggregate, not how it is assembled.

#349 covers the browser side of these three surfaces.

<details><summary><b>Use of AI Tools</b></summary>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

</details>

jigneshbhavani:
Fixes #137

Both Network Admin render callbacks returned network-wide presence to any caller, since only the column registration was gated. That holds while ours is the only thing that can put a column called `presence_online` on those screens, which is not a property this plugin controls. Called as a subscriber:

| Uncapped call | main | this branch |
| --- | --- | --- |
| `wp_presence_render_network_users_column( '', 'presence_online', $editor )` | `localhost:8891/testpath0000025/` | returns `$output` untouched |
| `wp_presence_render_network_sites_column( 'presence_online', $blog_id )` | avatar stack and count | nothing |

Notes:
- Gated rather than pinned as unreachable with a test, which was the open question on the issue. Two lines, it matches every other entry point in those files, and unreachability stops being true the moment another plugin registers the name.
- The cross-site denial the issue asks for already holds, so that test is a regression guard rather than a fix. Same for the site-scoped read.
- `test_users_column_shows_a_dash_for_an_offline_user` called the renderer uncapped and passed because the user was offline. It becomes a network admin first, which is what it always meant to assert.
- The `wp_presence_network_capability` docblock now says what a lowered value grants.

Testing:
- `npm run test:multisite`: 466 tests, 986 assertions, 3 skipped.
- `npm test`: 466 tests, 739 assertions, 131 skipped.
- `composer phpcs`: clean.
- Mutations: caps read from the main site inside the switch fails the denial test, the read pinned to `$wpdb->base_prefix` fails the isolation test, deleting either gate fails that column's test, and nothing else moves.
- Browser: an mu-plugin filtering the capability to `do_not_allow` and registering its own `presence_online` column, which is the third-plugin case above. On `main` that viewer sees the avatar stacks and the site names; on this branch both columns are empty. A super admin sees no change either way.

<details>

AI assistance: Yes  
Tool(s): Claude Code  
Model(s): Claude Opus 5  
Used for: Drafting the gate, the tests and this description. I ran the measurements, the mutation checks and the browser comparison, and reviewed what landed.

</details>

### DIFF
--- a/includes/network-functions.php
+++ b/includes/network-functions.php
@@ -523,6 +523,12 @@ function wp_presence_network_capability() {
 	/**
 	 * Filters the capability required to see network-wide presence data.
 	 *
+	 * What this guards is network-wide: who is online on every site, and which
+	 * sites a given user is online on. Lowering it to a capability a site-level
+	 * role already holds, such as 'edit_posts', hands presence for the whole
+	 * network to every holder on every site, including sites they are not a
+	 * member of.
+	 *
 	 * @param string $capability Default 'manage_network'.
 	 */
 	return apply_filters( 'wp_presence_network_capability', 'manage_network' );
--- a/includes/network-sites-list.php
+++ b/includes/network-sites-list.php
@@ -86,11 +86,19 @@ function wp_presence_network_aggregation_notice() {
  * snapshot is read once for the request; what each row adds is resolving the
  * handful of people whose avatars it is about to draw.
  *
+ * Gates on the capability again rather than trusting the registration above:
+ * core calls this for whatever columns the screen ended up with, and ours is
+ * not the only thing that can put a name in that list.
+ *
  * @param string $column_name Column being rendered.
  * @param int    $blog_id     The site ID for the current row.
  */
 function wp_presence_render_network_sites_column( $column_name, $blog_id ) {
 	if ( 'presence_online' !== $column_name ) {
+		return;
+	}
+
+	if ( ! current_user_can( wp_presence_network_capability() ) ) {
 		return;
 	}
 
--- a/includes/network-user-list.php
+++ b/includes/network-user-list.php
@@ -108,6 +108,10 @@ function wp_presence_register_network_users_column( $columns ) {
  * avatar URL for everyone online would be the whole cost of the read for
  * output this column never uses.
  *
+ * Gates on the capability again rather than trusting the registration above:
+ * core calls this for whatever columns the screen ended up with, and ours is
+ * not the only thing that can put a name in that list.
+ *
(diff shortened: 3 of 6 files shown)


josephfusco (approved):
LGTM, good add!

Issue #413, opened 2026-08-29
PHPStan analyses against WordPress 6.9 stubs while the suite runs on 7.1


josephfusco:
PHPStan analyses against WordPress 6.9 while the suite runs on 7.1.

```json
// composer.json:13
"php-stubs/wordpress-stubs": "^6.9",
```

A caret on `6.9` resolves `>=6.9 <7.0`, so the constraint cannot reach 7.x. Run 33254997943 installed `php-stubs/wordpress-stubs (v6.9.4)`.

## Evidence

| | Version |
| --- | --- |
| wp-env, both containers | 7.1 |
| PHPStan stubs | 6.9.4 |
| Newest published stubs | `v7.1.0` |

Anything core added in 7.0 or 7.1 reads as an undefined function to PHPStan, and any signature that changed across those two majors is checked against the old one. Neither shows up as a failure, which is the problem: the analysis passes by not knowing.

## Tasks

- [ ] Move the constraint to the 7.x line
- [ ] Check whether the bump surfaces findings that the 6.9 stubs were hiding

## Worth knowing

The stubs track releases, so this needs bumping every major. Whatever fixes #412 is the natural place to notice it, since both are the same shape: a version claim nothing verifies.

<details>
<summary><b>Use of AI Tools</b></summary>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

Used for: Assisting with the audit of the analysis toolchain

</details>

i-am-chitti:
@josephfusco As I've picked 412, I'll work on this too.

i-am-chitti:
The fix is not working. `composer install` fails, because a second stubs package blocks it. Both are required together in [`composer.json:13-14`](https://github.com/WordPress/presence-api/blob/f59f6be8d1ab776c219eff04ded5c50d0bfe6a4a/composer.json#L13-L14):

    php-stubs/wp-cli-stubs v2.12.0 requires php-stubs/wordpress-stubs ^4.7 || ^5.0 || ^6.0 -> found
    php-stubs/wordpress-stubs[v4.7.14, ..., v6.9.4] but it conflicts with your root composer.json require (^7.1).

So line 13 can't change unless line 14 changes too. Upstream already fixed this in php-stubs/wp-cli-stubs#24 (merged May 2026), but the newest tag is still v2.12.0 from June 2025 (11 months gap), so the fix was never released.

[...]

PR #477, opened 2026-09-08
fix: stop PHPStan's bootstrap from silently exiting before analysis

josephfusco:
Related: #413

@i-am-chitti flagged this while investigating #413; the stub-version bump discussed there is still blocked upstream on wp-cli-stubs.

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Sonnet 5
Used for: Assisting with diagnosis and the fix.

</details>

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -1,7 +1,7 @@
 parameters:
 	level: 3
 	bootstrapFiles:
-		- presence-api.php
+		- phpstan-bootstrap.php
 	paths:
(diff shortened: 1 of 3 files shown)


phpstan.neon.dist
i-am-chitti: dead config - `tests/e2e/demo-seeder.php`

i-am-chitti (approved):
Working as expected. Verified on local, PHPStan is analyzing statically.
LGTM 🚀

PR #492, opened 2026-09-09
feat: add a Settings link to the plugin row actions

### LINKED ISSUE #484: Add a Settings link to the plugin row actions
`wp_presence_plugin_action_links()` (`presence-api.php:292`) adds one link, "View Online Users", to the plugin's row on the Plugins screen. The recording toggle from #400 and #441 has no link there, even though it's just a field on Settings > General.

`tests/test-plugin-action-links.php` hardcodes the current single-link count in two places, so a correct implementation fails tests that look unrelated at first glance:

- `test_prepends_a_single_link()` asserts the result has exactly 2 entries
- `test_returns_only_our_link_when_no_links_exist()` asserts exactly 1

Both need their counts updated. The other six tests key off `$links[0]` as the online users link specifically, so keep that link first; add Settings after it rather than before.

## Tasks

- [ ] Add a "Settings" link to `wp_presence_plugin_action_links()`, pointing at `options-general.php`, added after the existing online users link
- [ ] Update the two link-count assertions in `tests/test-plugin-action-links.php`
- [ ] Add a test asserting the Settings link's href and label

## Test

Activate the plugin and check its row on `wp-admin/plugins.php`. Both links should show, online users first.

Part of #483.

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Sonnet 5
Used for: Assisting with splitting #483 into a scoped sub-issue, and verifying which existing tests break by making the change in a scratch branch.

</details>

guzel:
Beyond the issue's task list, this also renames `test_prepends_a_single_link` and `test_returns_only_our_link_when_no_links_exist` — both now assert the opposite of what their names say. Happy to drop the renames if you'd rather keep the diff to the three tasks.

The Settings link points at `options-general.php` as specified, which lands on General and leaves you to find the "Presence" row. The field registers `label_for => wp_presence_recording`, so `options-general.php#wp_presence_recording` would jump straight to it — say the word if you'd prefer that.

Fixes #484

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Opus 5
Used for: Drafting the implementation and tests; reviewed by me.

</details>

### DIFF
--- a/presence-api.php
+++ b/presence-api.php
@@ -283,11 +283,12 @@ function wp_presence_deactivate( $network_wide = false ) {
 /**
  * Adds action links to the plugin list table.
  *
- * The plugin has no settings screen, so the link points at the Users list
- * filtered to the users who are currently online.
+ * The online users link points at the Users list filtered to the users who are
+ * currently online. The plugin has no settings screen of its own, so the
+ * settings link points at Settings > General, where the recording switch lives.
  *
  * @param string[] $links Existing plugin action links.
- * @return string[] Action links with the online users link prepended.
+ * @return string[] Action links with the plugin's own links prepended.
  */
 function wp_presence_plugin_action_links( $links ) {
 	$online_users_link = sprintf(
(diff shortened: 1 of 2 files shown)


josephfusco (approved):
LGTM, nice first contribution!

Issue #498, opened 2026-09-10
e2e's ~8 minute wall-clock is bounded by wp-env boot cost, not test count


josephfusco:
e2e takes ~8m18s on a recent run: wp-env boot (single ~1m54s + multisite ~35s), setup/cleanup (~1m22s), and the Playwright run itself (~4m27s, serial, `workers: 1` in playwright.config.js, no comment explaining why).

Sharding won't fix this: each shard pays its own wp-env boot independently, so wall-clock can't drop below that ~3m51s fixed cost no matter how many shards run, while billed CI minutes keep climbing. The boot cost itself has to shrink.

## Open questions

- Do the specs need `workers: 1`, or is that a stale default? They share one live wp-env/database, so parallelizing risks real races between specs, not just a config flip.
- Where does the ~1m54s single-site boot go (image pull, WP install, plugin activation), and is any of it cacheable?

josephfusco:
Closing as won't-fix. PR #500 explored caching wp-env's WordPress core checkout: the wrong-path bug was real, but making the cache actually work (pinning WP_ENV_HOME) restores stale host state across runs and broke the multisite wp-env instance in testing. Max upside was ~30s off the ~8m24s job.

Also looked at gating the multisite instance behind a path filter for a bigger (~2min) win, but shared code between single-site and multisite paths means a safe filter would rarely skip anything on real changes.

The workers: 1 question is answered: specs share one live wp-env db and some truncate wp_presence or delete all posts/users, so parallelizing would race.

PR #500, opened 2026-09-11
ci: cache wp-env's WordPress core checkout in the e2e job

josephfusco:
Related: #498

Opened as a draft since the wp_presence table truncation there in presence-screenshots.test.js, plus the deleteAllPosts/deleteAllUsers calls elsewhere, made me draw the workers:1 conclusion from reading the specs rather than from a real racy-parallel CI run, and I haven't yet measured actual before/after wall-clock on a same-day cache hit in CI.

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Sonnet 5
Used for: Assisting with investigating wp-env's boot cost and drafting the fix.

</details>

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -16,6 +16,10 @@ jobs:
     name: Playwright
     runs-on: ubuntu-24.04
     timeout-minutes: 30
+    env:
+      # Pinned so the cache path below doesn't depend on wp-env's runner-specific
+      # snap detection (~/.wp-env vs ~/wp-env).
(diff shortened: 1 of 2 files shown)


josephfusco:
Closing as won't-fix. Confirmed the wp-env cache path bug was real and fixable, but caching wp-env's home directory restores stale host-side state across runs and broke the multisite instance in testing (wp_blogs missing, site not found) on the first warm-cache run. The capped upside (~30s off an 8m24s job) doesn't justify that risk.

PR #514, opened 2026-09-21
perf: keep the collaboration edge state in the presence table

### LINKED ISSUE #512: Collaboration edge state writes to `wp_options` on hosts without a persistent object cache
## Finding

Without a persistent object cache a transient is a `wp_options` row. `wp_presence_check_collaboration_threshold()` rewrites its transient every tick while a room holds 2+ editors, so an uncached site takes an option write per heartbeat per collaborating room.

[`includes/heartbeat.php:397-440`](https://github.com/josephfusco/presence-api/blob/fd22a7670149e83b1c400c6d0276a0201cf7b518/includes/heartbeat.php#L397-L440)

```php
$stored     = get_transient( $key );
$prev_count = false === $stored ? 1 : (int) $stored;
...
set_transient( $key, $editor_count, wp_presence_get_timeout( WP_PRESENCE_DEFAULT_TTL ) );
```

A pair on a 15 second heartbeat refreshes the timeout option eight times a minute per room. It is the plugin's only transient.

## Proposal

Park the previous count in `wp_presence`, in the room it describes, under a reserved `client_id`.

| `room` | `client_id` | `user_id` | `data` |
| --- | --- | --- | --- |
| `postType/post:42` | `_collab` | `0` | `{"count":2}` |

Read is free: `wp_get_presence( $room )` at line 398 already returns every live row in the room. Write goes through `wp_set_presence()`, so the row ages out on the presence TTL and inherits the unchanged-write skip in `wp_presence_write_is_redundant()`. `wp_presence_count_editors()` filters on `editor-`, so the count is unaffected.

Net: one upsert instead of one `set_transient()`, no extra read.

## Cost

Room-scoped reads must exclude reserved `client_id`s or show a phantom participant.

| Path | Surface |
| --- | --- |
| `includes/functions.php:66` (`wp_get_presence()`) | Who's Online, post list, avatar stacks |
| `includes/functions.php:600` (`wp_get_presence_by_room_prefix()`) | Room listings |
| `includes/functions.php:648` (`wp_get_presence_summary()`) | Entries per room |
| `includes/functions.php:957`, `:993`, `:1053` (`wp_presence_hydrate_room_users()`) | `COUNT(DISTINCT user_id)` counts `0` as a user |
| `includes/class-wp-rest-presence-controller.php:332`, `:342` | Room total and paginated entries |

`db-viewer.php:38` and the CLI listing show raw rows, so it belongs there. `privacy.php:88` filters on `user_id`.

## Tasks

- [x] Reserve the `_` prefix for non-client rows; document under README Client IDs
- [x] Write the count as a `_collab` row, read it from the entries already returned
- [x] Filter reserved rows from the paths above
- [x] Delete the row below two editors
- [x] Reject reserved `client_id`s on the REST create and delete routes
- [x] Drop `wp_presence_collaboration_state_key()`
- [x] Tests: no reserved row in REST, Who's Online, or an avatar stack

Two items came off the list in #514:

| Dropped | Reason |
| --- | --- |
| Clear stale `_transient_wp_presence_collab_*` options on upgrade | `wp-admin/admin.php` schedules a daily `delete_expired_transients` event, which deletes the `_transient_` row with its `_transient_timeout_` pair once the timeout has passed. Every leftover here is already expired. |
| Tests: edges hold with `wp_using_ext_object_cache()` false | No transient is left to read, so nothing on this path varies with the object cache. |

The write goes through `wp_presence_write_row()`. `wp_set_presence()` would re-run the redundant-write check over a second `SELECT` that the room read has already answered.

## Alternatives

| Approach | Why not |
| --- | --- |
| Own room, `_state/postType/post:42` | No read filtering, but a second SELECT per tick and it surfaces in the summary and prefix reads |
| Count on each editor's own row | No new row, no filtering, closes the both-fire race, but makes `wp_presence_collaboration_started` per participant instead of per room |

josephfusco:
Fixes #512

Without a persistent object cache this was an option write per heartbeat per
collaborating room, and the plugin's only transient.

Two deviations from the issue:

- `wp_get_active_rooms()` needed filtering as well. It is not in the issue's
  cost table, and the REST rooms endpoint reads it with hydration off, where
  the reserved row counts as a user.
- Stale `_transient_wp_presence_collab_*` options are left to core. They expire
  on the presence TTL and `delete_expired_transients` sweeps them daily, so the
  upgrade routine the issue asks for would duplicate core.

<details>

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5

Used for: Assisting with implementation and testing

</details>


includes/heartbeat.php
```diff
@@ -427,14 +443,38 @@ function wp_presence_check_collaboration_threshold( $room ) {
...
+	$age       = $stored ? time() - (int) strtotime( $stored->date_gmt . ' UTC' ) : 0;
+
+	// Same rule as wp_presence_write_is_redundant(), decided from the row this
+	// request has already read rather than from a second SELECT.
+	if ( $unchanged && $threshold > 0 && $age <= $threshold ) {
+		return;
+	}
+
+	wp_set_presence( $room, wp_presence_collaboration_state_client_id(), array( 'count' => $count ) );
+}
```
i-am-chitti, line 488: This function checks if the write is redundant using `$stored`, then calls `wp_set_presence()` which [runs the same redundancy check again](https://github.com/WordPress/presence-api/blob/52919c8151540530e15e703ed5ff0e80c677d134/includes/functions.php#L452) with a fresh `SELECT`. So we still do an extra DB query on every write, which is what this PR was meant to avoid.

josephfusco: Thanks. The upsert is now a private `wp_presence_write_row()`, so the state write skips that SELECT. Covered by a test that fails on the old code.

